PR #17, opened 2019-11-29
Set up cpp test framework

hsautessella:
A C++ testing framework has been written. 

One test has been written for `IIget_ascii_file.cpp` as an example and proof of concept. Tests can be run using ctest.

This is a similar setup as is implemented in Horace

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ find_package(MPI REQUIRED)
 include(herbert_FindMatlab)
 if(${BUILD_TESTS})
   include(herbert_FindGTest)
+  enable_testing()
 endif()
 
 # Set the destination of our build artifacts
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ set(CMAKE_CXX_STANDARD 11)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 # Set our options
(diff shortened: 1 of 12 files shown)


nickbattam-tessella:
How complex would it be to move the duplicated pieces of this from Horace+Herbert into a new shared Common repo? I'm thinking particularly about the cmake 'add unit test' method and the environment.h/cpp utility function?

abuts:
It would not be too difficult, but we have discussed this issue and decidied agains it. Mainly because of Toby's insistance.

abuts:
Curently Herbert should not depend on Horace, but Horace is impossible without Herbert -- this is dependence. 
And I think, G-gests should sit in Herbert

nickbattam-tessella:
Having a common module doesn't force herbert to depend on horace, i.e.
```
   horace <---- herbert
      ^             ^
       \           /
           common 
```

abuts:
This will be very difficult as Herbert is actually a part of Horace. Very few Herbert functions are not used in Horace.

nickbattam-tessella:
@hsautessella what is blocking this PR being completed?

abuts:
my commit to this PR is erroneous. Actually its not the commit to PR, its just wront ticket number mentioned.

hsautessella:
I think it's good to go, we were just having the discussion about dealing with the similar code that is in Herbert. But this is fine to go in for now

PR #19, opened 2019-12-02
Fix @oriented_lattice/bmatrix typo `arly`->`arlu`

hsautessella:
The second return value in `bmatrix` was not being set correctly due to a typo.

### DIFF
--- a/herbert_core/classes/instrument_classes/@oriented_lattice/bmatrix.m
+++ b/herbert_core/classes/instrument_classes/@oriented_lattice/bmatrix.m
@@ -33,4 +33,4 @@ function [b, arlu, angrlu] = bmatrix(obj)
 
 angdeg = obj.angdeg;
 alatt  = obj.alatt;
-[b,arly,angrlu] = bmatrix(angdeg,alatt);
+[b, arlu, angrlu] = bmatrix(angdeg, alatt);


nickbattam-tessella:
are any of the calling methods using this third return value and so will see a change in behaviour following the fix??

abuts:
There are no unit tests for this particular function, as such (shame) but multiple tests for the oriented_lattice exist and there is some Toby's test which I've moved to Horace, as it depends on Horace functions uses b-matx istelf

nickbattam-tessella:
Since the third return value would have been (historically) unset is that value useful to expose? People can't have been using it or the error would have been discovered.

abuts:
There were two definitions of b-matrix in Herbert and Horace -- the first was internal for oriented lattice. 
I've just moved b-matrix definition from Horace to Herbert and changed internal Herbert definition to refer to public b-matrix. Made typo Harry have fixed.

nickbattam-tessella:
Adding unit tests would be a nice-to-have addition to the PR ([Campsite Rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html))

hsautessella:
Looks like there are some tests, but they are failing for different reasons. https://github.com/pace-neutrons/Herbert/blob/master/_test/test_IX_classes/test_oriented_lattice.m#L187

```
===== Test Case Failure =====
Location: 
Name:     test_oriented_lattice::test_matrixes

C:\Users\ejo73213\PACE\Herbert\_test\test_IX_classes\test_oriented_lattice.m at line 187

Input elements are not all equal within absolute tolerance: 1e-09

First input:
    4.6189   -2.3093   -2.3093
         0    4.0002   -3.9996
         0         0    0.0698

Second input:
    6.2832         0         0
         0    6.2832         0
         0         0    6.2832
```

nickbattam-tessella:
Can you break `test_matricies` up into smaller test-one-thing tests; it might help us see whether there's a pattern of failure.

abuts:
The test for ub matrix does not look good at all. As its my unfinished buiziness, I suspect I should pick this thing up.

abuts:
It may be more then just a typo

PR #20, opened 2019-12-02
16 code relocation

abuts:
this completes the code relocation and fixes the oriented lattice unit test failure as another typo have been found there.


herbert_core/classes/instrument_classes/@oriented_lattice/bmatrix.m
```diff
@@ -33,4 +33,4 @@
 
 angdeg = obj.angdeg;
 alatt  = obj.alatt;
-[b, arlu, angrlu] = bmatrix(angdeg, alatt);
+[b,arly,angrlu] = bmatrix(alatt,angdeg);
```
hsautessella: Looks like we're reverting a fix on the second return argument here

hsautessella:
```suggestion
[b,arlu,angrlu] = bmatrix(alatt,angdeg);
```

PR #23, opened 2019-12-09
22 simulink to herbert

abuts:
The task appeared to be much simple as `which` command in Matlab apparently returns realpath. 
This is small changes I had to do doing release. 
Similar changes are in Horace and this request is pared with the Horace one.

### DIFF
--- a/herbert_core/configuration/@herbert_config/private/process_unit_test_path.m
+++ b/herbert_core/configuration/@herbert_config/private/process_unit_test_path.m
@@ -9,6 +9,13 @@ if isempty(rootpath)
         'herbert_init is not on Matlab search path');
 end
 tests_path = fullfile(rootpath,'_test');
+if ~(exist(test_path,'dir')==7)
+    warning('HERBERT_INIT:invalid_setup',...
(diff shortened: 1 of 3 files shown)


herbert_core/configuration/@herbert_config/private/process_unit_test_path.m
```diff
@@ -9,6 +9,13 @@
         'herbert_init is not on Matlab search path');
 end
 tests_path = fullfile(rootpath,'_test');
+if ~(exist(test_path,'dir')==7)
```
nickbattam-tessella: what is this magic `7`?

abuts: That's Matlab code for directory. file would be 2

abuts: Matlab is relatively old languge, so has number of such gems lying around.

nickbattam-tessella: it would be useful to put that in as comment for future record

Issue #24, opened 2019-12-13
Bug in parallel configuration if worker is not availabe


abuts:
The bug has been reported by user who recently installed Horace but does not have worker available and not intending to use parallel extensions. 
The installation fails. 
This should be fixed as some users may not wont parallel extensions.

PR #25, opened 2019-12-16
24 no parallel

abuts:
This fixes the issue #24, which user experience when his installation does not have parallel framework enabled.


_test/test_mpi/test_iMPI_methods.m
```diff
@@ -21,9 +25,21 @@
             obj.current_config_folder = cs.config_folder;
             
             pc = parallel_config;
+            if strcmpi(pc.parallel_framework,'n/a')
```
nickbattam-tessella: is `n/a` a good constant with a 'special' character in? would `none` be simpler? It's used as a hard-coded string in a *lot* of places

_test/matlab_xunit_ISISextras/@TestPerformance/TestPerformance.m
```diff
@@ -270,6 +270,7 @@
             else
                 test_name = [comp_name,'_',framework_name];
             end
+            test_name = regexprep(test_name,'[/\\]','_');
```
nickbattam-tessella: if you don't have `n/a` as the name this conversion isn't required

abuts:
actually none is good, I thought about NaN which is not

PR #27, opened 2019-12-17
26 cluster mpi on unix

abuts:
Main time on this ticket was spent working on the problem why compiled cpp communicator is not working. And after it become working, I was able to build it using cmake. 

Added Matlab compilation procedure to use for speed and convenience until Cmake is as reliable.


herbert_core/classes/MPIFramework/@ClusterMPI/private/check_mpi_mpiexec_can_be_enabled_.m
```diff
@@ -1,18 +1,20 @@
...
-mess = [];
-if isempty('cpp_communicator')
-    ok = false;
+
+if isempty(which('cpp_communicator'))
     mess = 'Can not find cpp_communicator mex file on Matlab search path';
+    error('PARALLEL_CONFIG:not_available',mess);
+end
+try
+    ver = cpp_communicator();
```
nickbattam-tessella: what is `ver` used for here?

admin/herbert_mex_mpi.m
```diff
@@ -12,13 +12,14 @@
 input_files = cellfun(@(fn)fullfile(code_folder,fn),input_files,'UniformOutput',false);
 
 add_include = ['-I',mpich_folder,'include'];
-add_lib = ['-L',mpich_folder,'lib'];
+mpich_lib = fullfile(mpich_folder,'lib');
+add_lib =['-L',mpich_lib ]; %,['--rpath-link=',mpich_lib]};
```
nickbattam-tessella: the dead code at the end of this line can be deleted

admin/herbert_mex_mpi.m
```diff
@@ -12,13 +12,14 @@
 input_files = cellfun(@(fn)fullfile(code_folder,fn),input_files,'UniformOutput',false);
 
 add_include = ['-I',mpich_folder,'include'];
-add_lib = ['-L',mpich_folder,'lib'];
+mpich_lib = fullfile(mpich_folder,'lib');
```
nickbattam-tessella: defining this as '/usr/local/mpich/lib' does this work on Windows?

herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
```diff
@@ -201,7 +201,11 @@ function check_availability(obj)
                 % future.
                 mpi_exec = fullfile(rootpath,'DLL','_PCWIN64','MS_MPI_R2019b','mpiexec.exe');
             else
+                % use system-defined mpiexec
```
nickbattam-tessella: you're handling this as PC (I assume Windows) and non-PC ... does the same non-PC configuration work on MacOS, Linux and SCARF deployments?

herbert_core/classes/MPIFramework/@ClusterMPI/private/check_mpi_mpiexec_can_be_enabled_.m
abuts: ver is not used, but mex file interface, returning ver needs this kind of call -- one arg-out and no arg-in, so have to be here.

herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
abuts: that's Linux build and Mac should be the same. Scarf too if mpiexec is exposed to users. But I think we will need another kind of Cluster, wrapping around qsub or bsub or whatever.

admin/herbert_mex_mpi.m
abuts: Have doubts if this script is appropriate at all in favour of cmake. But after morinign discussion will change it properly.

herbert_core/classes/MPIFramework/@ClusterMPI/private/check_mpi_mpiexec_can_be_enabled_.m
nickbattam-tessella: Doesn't Matlab support dummy argument/return values (`~`)? Having unused variables declared makes the code more confusing

abuts: Matlab supports, but I am not sure how mex would recognize it -- as an argument or no argument. I suspect as no argument, because it should not generate output value then. 
needs more investigations...
Check within custom mex return version in case of one output and no input arguments and I do not think its worth chaning this.

PR #30, opened 2019-12-23
Jenkins build pipeline for scientific linux

nickbattam-tessella:
Initial test at this stage; do not merge.


hsautessella:
I think this is at a similar state as it is in Horace and the build is passing. I'm happy to merge this if you are @abuts?

PR #32, opened 2020-01-06
Introduce 'CapSize' to custom_errorbars

g5t:
The errorbar cap length was introduced as an optional named value 'CapSize', with units in points, in MATLAB 2016b.
Setting the cap length to zero has the effect that very-short errorbars (compared to the vertical axis) are not drawn, compared to the old method of drawing the symbols and errorbars separately which draws a dot in the same situation.

This function can be simplified if the minimum version of MATLAB supported by Herbert is ever increased to 2016b. (Efficient use of Euphonic and/or brille by Horace requires support for Python `memoryview` objects, which was introduced in MATLAB 2018b.)


abuts:
Well, such changes can be merged by developer himself.

PR #34, opened 2020-01-07
Add post build commands for mex libs

hsautessella:
This PR writes a `herbert_add_mex` function in CMake. This function basically does the same thing as CMake's built-in `matlab_add_mex` except it will copy the mex library to a given location after compilation. The default copy location is `herbert_core/DLL`


abuts:
As I will probably use it now and it makes sence, I would merge it without exptensive external checks

Issue #28, opened 2019-12-18
Assync send-receive handler


abuts:
This is Herbert implementation of the https://github.com/pace-neutrons/Horace/issues/70

abuts:
This one is completed

PR #35, opened 2020-01-07
28 assync send receive

abuts:
This is main part of the resolution on the ticket #28 
This thing is  not entirely completed but have reasonable unit tests coverage so I am making PR as the changes are already too big to understand easily. 

the thing remain for all tests to run would be proper serialization of MException class but I will do it as a separate ticket as it is clear separate job. 

Otherwise, this PR adds asynchroneous message exchange using cpp_communicator class and the message exchange indeed happens. 

To finish cpp_communicator one needs synchroneous message exchange but this will be done within a separate ticket.


_LowLevelCode/CPP/test/cpp_communicator.tests/cpp_communicator_test.cpp
```diff
@@ -0,0 +1,453 @@
...
+#include "utility/environment.h"
+
+#include <gtest/gtest.h>
+
+#include <vector>
+
+using namespace Herbert::Utility;
+
+
+TEST(TestCPPCommunicator, send_assynchroneous) {
```
hsautessella: Could we try and make these test names more descriptive? The format could be something like `this_does_this_given_this`, it makes it much easier to determine what the test is actually doing.

_test/test_mpi/job_dispatcher_common_tests.m
```diff
@@ -81,8 +81,8 @@ function my_delete(varargin)
             assertEqual(numel(outputs),3);
             assertTrue(isa(outputs{1},'MException'));
             assertTrue(isa(outputs{2},'MException'));
-            assertTrue(isa(outputs{3},'MException')||...
-                (isstruct(outputs{3})&&isfield(outputs{3},'time')));
+%             assertTrue(isa(outputs{3},'MException')||...
+%                 (isstruct(outputs{3})&&isfield(outputs{3},'time')));
```
hsautessella: If this isn't used, should we just remove it?

_LowLevelCode/CPP/test/cpp_communicator.tests/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+    "${CXX_SOURCE_DIR}/utility/environment.cpp"
+)
+
+set(HDR_FILES
+    "${CXX_SOURCE_DIR}/cpp_communicator/cpp_communicator.h"
+    "${CXX_SOURCE_DIR}/cpp_communicator/input_parser.h"
+    "${CXX_SOURCE_DIR}/cpp_communicator/MPI_wrapper.h"
+    "${CXX_SOURCE_DIR}/utility/environment.h"
+)
+include_directories(${MPI_CXX_HEADER_DIR})
```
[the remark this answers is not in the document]
abuts: The tests which involve exception are failing as it does not serialize properly. The next ticket I am doing should fix this issue.

_test/test_mpi/job_dispatcher_common_tests.m
abuts: This is difficult question. Different runs in different combination and different workers speed return different result. I am keen to catch them all and was trying to achieve that, but apparently failed. Looking to better ways of catching different type of returns.

_LowLevelCode/CPP/test/cpp_communicator.tests/cpp_communicator_test.cpp
abuts: generally agree but its exactly what this test is doing -- testing asynchroneous send given nothing. Is it not clear?
CPPCommunicator_sends_asyncrhoneous_ (no conditions)

_LowLevelCode/CPP/test/cpp_communicator.tests/cpp_communicator_test.cpp
```diff
@@ -0,0 +1,453 @@
...
+    // Mark all messages as delivered
+    auto MessCache = wrap.get_async_queue();
+
+    auto last = MessCache->rbegin();
+    last->theRequest = 1; // mark last message delivered;
+    wrap.labSend(5, 5, false, &test_mess[0], test_mess.size());
+    ASSERT_EQ(4, wrap.assync_queue_len());
+
+
+    for (auto it = MessCache->begin(); it != MessCache->end(); it++) {
```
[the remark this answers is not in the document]
abuts: will it still be iterator? not familiar with this form very well....

_LowLevelCode/CPP/test/cpp_communicator.tests/CMakeLists.txt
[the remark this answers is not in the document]
hsautessella: `target_include_directories(${target} PRIVATE ${MPI_CXX_HEADER_DIR})` essentially just adds `-I /opt/mpi/include/` (or wherever the MPI include directory is) to the `g++` command for that target. 

Instead of `PRIVATE` you can specify `PUBLIC`. `PUBLIC` means that the headers will be part of the public interface of whatever you're building, private means you cannot access those headers through the library you're buillding. https://stackoverflow.com/questions/26243169/cmake-target-include-directories-meaning-of-scope

Using `include_directories` will add `-I /opt/mpi/include/` to __all__ targets that CMake builds, which we obviously don't want.

_LowLevelCode/CPP/test/cpp_communicator.tests/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+#
+set(LIBS
+    "${Matlab_UT_LIBRARY}"
+)
```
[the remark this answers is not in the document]
abuts: Done. BTW, to run tests on Windows I need to modify VS project adding PATH=c:\programming\Matlab2018b64\bin\win64;
$(LocalDebuggerEnvironment) to CPP_communicator_tests in project->properties->ConfigurationProperties->Debugging->Environment. (path to libmex.dll) Should not CMake do that?

_test/test_mpi/job_dispatcher_common_tests.m
abuts: Addressed in next ticket where these forms are higly modified (and should be modified further)

_LowLevelCode/CPP/test/cpp_communicator.tests/cpp_communicator_test.cpp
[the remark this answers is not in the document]
abuts: still do not understand what is the difference between `auto &message` and `auto &&message` so will be very reluctant to use this form in perfofmane critical peaces but looks fine here

_LowLevelCode/CPP/test/cpp_communicator.tests/CMakeLists.txt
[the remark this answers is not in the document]
hsautessella: Yes, the Matlab bin is added to the path in CMake, this is done in the `herbert_add_unit_test` CMake function. But this only works if you're running the tests with CTest. If you're running the test executables directly, you'll need to add the Matlab bin to the path yourself.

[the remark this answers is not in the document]
abuts: OK than, this works fine in this case.

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
```diff
@@ -1,25 +1,114 @@
...
+        return this->assyncMessList.size();
+    }
+    // the tag of message, containing data (processed differently, not yet implemented.)
+    static int data_mess_tag;
 
-	~MPI_wrapper() {
-		this->close();
-	}
+    //----------------------------------------------------------------------------------
+    // The methods used in unit tests -- have no meaning in real communications
```
nickbattam-tessella: if they're only used in unit tests should the be part of the PUBLIC API?

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
```diff
@@ -1,25 +1,114 @@
...
+            return true;
+        for (auto it = assyncMessList.rbegin(); it != assyncMessList.rend(); it++) {
+            if (it->theRequest == 0) {
+                return true;
+            }
+        }
+        return false;
+    }
+    // get first not "delivered" message in test mode.
+    SendMessHolder* get__message() {
```
nickbattam-tessella: why does this have double underscore?

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
```diff
@@ -1,25 +1,114 @@
...
+        }
+        return nullptr;
+    }
 
-	int labIndex;  // index of the current MPI lab (worker)
-	int numProcs; // total  number of MPI labs (workers)
 private:
+    // the length of the queue to keep assynchroneous messages. If this length is exceeded,
+    // something is wrong and the job should be interrupted
+    int assynch_queue_max_len_;
```
nickbattam-tessella: spelling error throughout this file `async_queue_max_len__`  it's `asynchronous` not `assynchroneous` or `assyncroneous` and `async` not `aasync` or `aasynch`

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
```diff
@@ -44,16 +100,21 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
     //* Check and parce input  arguments. */
-    char *data_buffer(nullptr);
+    uint8_t * data_buffer(nullptr);
+
+    std::vector<int32_t> data_addresses;
+    std::vector<int32_t> data_tag;
+    bool is_synchroneous(false);
 
-    int data_address(0), data_tag(0);
+    AdditinalParamHolder AddPar;
```
nickbattam-tessella: spelling error `AdditionalParamHolder`

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
```diff
@@ -44,16 +100,21 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
     input_types work_type;
 
 
     class_handle<MPI_wrapper>* pCommunicatorHolder = parse_inputs(nlhs, nrhs, prhs,
-        work_type, data_address, data_tag, nbytes_to_transfer, data_buffer);
+        work_type, data_addresses, data_tag,is_synchroneous,
+        data_buffer, nbytes_to_transfer, AddPar);
 
     // avoid problem with multiple finalization
     if (pCommunicatorHolder == nullptr) { // this can happen only if close_mpi is selected and the framework had been already finalized
```
nickbattam-tessella: missing brackets around `if (nlhs>0) { body }`

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
```diff
@@ -81,14 +149,33 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
+            int32_t* pAddrTag = (int32_t*)mxGetData(plhs[(int)labProbe_Out::addr_tag_array]);
+            for(size_t i=0;i< n_present;i++){
+                pAddrTag[2 * i + 0] = address_present[i]+1;
+                pAddrTag[2 * i + 1] = tag_present[i];
+            }
+        }
         break;
     }
+    case(clearAll): { // receive and discard all messages, directed to the framework
+        pCommunicatorHolder->class_ptr->clearAll();
```
nickbattam-tessella: no `break` -- should not rely on fallthrough

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
```diff
@@ -100,23 +187,25 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
     } // end case
     } // end switch
 
-     if (nlhs > 0)
-         plhs[(int)labIndex_Out::comm_ptr] = pCommunicatorHolder->export_hanlder_toMatlab();
-
+    if (nlhs > 0)
+        plhs[(int)labIndex_Out::comm_ptr] = pCommunicatorHolder->export_hanlder_toMatlab();
```
nickbattam-tessella: missing `{ }`

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
```diff
@@ -67,7 +128,14 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
     switch (work_type)
     {
     case(init_mpi): { // Initialize MPI communications and return labIndex and numLabs
-        pCommunicatorHolder->class_ptr->init();
+        pCommunicatorHolder->class_ptr->init(false, AddPar.assynch_queue_length,AddPar.data_message_tag);
+        n_workers = pCommunicatorHolder->class_ptr->numProcs;
```
nickbattam-tessella: where is this `n_workers` being used?

_LowLevelCode/CPP/cpp_communicator/input_parser.cpp
```diff
@@ -43,16 +43,17 @@ size_t   get_byte_length(const char*err_id, const mxArray *prhs) {
         std::stringstream buf;
         buf << " The input for " << err_id << "contains unknown vector type\n";
         mexErrMsgIdAndTxt("MPI_MEX_COMMUNICATOR:invalid_argument", buf.str().c_str());
+        // never reached but for compiler not to complain
```
nickbattam-tessella: this comment isn't correct; in the default case that line will be reached.

_LowLevelCode/CPP/cpp_communicator/input_parser.cpp
```diff
@@ -113,71 +133,124 @@ prhs  --  array of pointers to right hand side parameters
 work_mode         -- retrieved IO operations mode.
 data_address      -- address of the node to communicate with
 data_tag          -- MPI messages tag
-nbytes_to_transfer-- number of bytes to transfer over mpi. 
+is_synchroneous   -- for send/receive operations, if the communication mode is synchroneous
```
nickbattam-tessella: spelling: synchronous

_LowLevelCode/CPP/cpp_communicator/input_parser.cpp
```diff
@@ -113,71 +133,124 @@ prhs  --  array of pointers to right hand side parameters
...
-            std::stringstream err;
-            err << " Init mode needs only 1 but got : " << nrhs << " input parameters";
-            throw_error("MPI_MEX_COMMUNICATOR:invalid_argument", err.str().c_str());
-        }
-        class_handle<MPI_wrapper> *pCommunicator = new class_handle<MPI_wrapper>();
         work_mode = init_mpi;
-        return pCommunicator;
+        return process_init_mode("Init", prhs, nrhs, AddPar);
+    }
+    else if (mex_mode.compare("init_test_mode") == 0) {
```
nickbattam-tessella: why is `mex_mode` a string not an enum given we're switching on it?

_test/test_mpi/MessagesCppMPI_test3.m
```diff
@@ -0,0 +1,25 @@
+classdef MessagesCppMPI_test3 < MessagesCppMPI
```
nickbattam-tessella: `_test3`??

_test/test_mpi/test_CPP_MPI_exchange.m
```diff
@@ -13,24 +13,226 @@
...
+            [ok,err_mess,message] = serverfbMPI.receive_message(1,'started');
+            assertEqual(ok,MESS_CODES.ok,['Error: ',err_mess]);
+            assertEqual(message.mess_name,'started')
+            assertEqual(numel(message.payload),3);
+            assertTrue(isempty(message.payload{1}));
+            assertEqual(message.payload{2},'a');
+            assertEqual(message.payload{3},'b');
+            
+        end
+        function test_SendReceive(obj)
```
nickbattam-tessella: is this one test or lots of tests?

_test/test_mpi/test_MESS_NAMES_factory.m
```diff
@@ -0,0 +1,68 @@
+classdef test_MESS_NAMES_factory< TestCase
+    %
+    % $Revision:: 833 ($Date:: 2019-10-24 20:46:09 +0100 (Thu, 24 Oct 2019) $)
```
nickbattam-tessella: what is this Revision string?

herbert_core/classes/MPIFramework/@MessagesCppMPI/private/init_framework_.m
```diff
@@ -31,21 +41,25 @@
...
 end
-[obj.mpi_framework_holder_,obj.task_id_,obj.numLabs_]= cpp_communicator('init');
-
-
-
-
+if test_mode
+    [obj.mpi_framework_holder_,obj.task_id_,obj.numLabs_]= ...
+        cpp_communicator('init_test_mode'...
+        ,obj.assync_messages_queue_length_,obj.data_message_tag_);
```
nickbattam-tessella: comma placement

herbert_core/classes/MPIFramework/@MessagesFilebased/MessagesFilebased.m
```diff
@@ -211,6 +256,21 @@ function clear_messages(obj)
         function val = get.time_to_fail(obj)
             val = obj.time_to_fail_ ;
         end
+        %
+        function is = is_job_canceled(obj)
+            % method verifies if job has been canceled
+            if ~exist(obj.mess_exchange_folder_,'dir')
```
nickbattam-tessella: can't this just be `isCancelled = ~exist(....)` ?

herbert_core/classes/MPIFramework/messages/MESS_NAMES.m
```diff
@@ -29,36 +111,38 @@
             % usage:
             % id = MESS_NAMES.mess_id('completed')
             % or
-            % ids = MESS_NAMES.mess_id('completed','running','started')
+            % ids = MESS_NAMES.mess_id('completed','log','started')
             %
             % where id-s is the array of message id-s (tags)
             %
+            name2code_map=MESS_NAMES.name_tag_maps();
```
nickbattam-tessella: '2'?!

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
abuts: Well, I can make it protected and expose it on tester class, but would it be an overkill? 
API is not that extensive and will not be much more complicated (I think two additional public methods are necessary i.e. serialize/deserialize) and everything is clear (hopefully) as it is. 

"Make an program interface a fully fool-proof and only full will use it...."

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
abuts: that's probably typo. Will remove it.

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
abuts: ok -- not usually payin attention to this kind of thins as non-native speaker but they are important to native speakers

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
abuts: Its the school in Mantid not using brackets in loops and ifs for single line expression
I was critical to it initially, but get used to it. 
not sure, should we request the brackets or not -- would agree to common desision

_LowLevelCode/CPP/cpp_communicator/input_parser.cpp
abuts: because enums in Matlab look horrible. And strings are pretty clear

_test/test_mpi/MessagesCppMPI_test3.m
abuts: 3 workers -- can be said better....

_test/test_mpi/test_MESS_NAMES_factory.m
abuts: that's the string which is currently updated during every static distribution creation. Its nice to see where the file came from when using continuoys integration

herbert_core/classes/MPIFramework/@MessagesCppMPI/private/init_framework_.m
abuts: fixed in the next ticket. Let's lieve it alone here to avoid merge conflicts.

herbert_core/classes/MPIFramework/@MessagesFilebased/MessagesFilebased.m
abuts: it can, but I believe its other code there. It will return on this but other check should follow if it does not.

herbert_core/classes/MPIFramework/messages/MESS_NAMES.m
abuts: name to(2) code map ? is it not a common acronym?

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
abuts: Actually it was tried and not used any more -- removed

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
abuts: done

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
abuts: done

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
abuts: let me leave it as it is until we decide on bracket policy

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
abuts: fixed

_test/test_mpi/test_MESS_NAMES_factory.m
nickbattam-tessella: But `833` doesn't match any Git SHA; Git gives you full line-by-line providence for any file.

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
abuts: That's the leftover from the previous implementation. Removed

herbert_core/classes/MPIFramework/messages/MESS_NAMES.m
nickbattam-tessella: it's not one I like -  `name_to_code` and `code_to_name` are clearer. We're not in FORTRAN and limited to 6 character names

_LowLevelCode/CPP/cpp_communicator/input_parser.cpp
abuts: yes, may be should be removed.

herbert_core/classes/MPIFramework/@MessagesFilebased/MessagesFilebased.m
nickbattam-tessella: you are evaluating a boolean to set a boolean variable; then using that boolean variable to conditionally execute a further chunk of code.
It could be simplified to:
```
if exist(....)
    mess = obj.probe_all('any','canceled');
    is = ~isempty(mess)
else
    is = false
end
```
or even
```
is = exist(...) && ~isempty(obj.probe_all('any','canceled'))
```

_LowLevelCode/CPP/cpp_communicator/input_parser.cpp
abuts: done

_test/test_mpi/MessagesCppMPI_test3.m
abuts: ok, will rename

_test/test_mpi/test_CPP_MPI_exchange.m
abuts: the keyword is "test_" this is how the framework identify its cases. At least that what I beliveve in

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
hsautessella: I'll +1 for always using brackets. It leads to fewer mistakes if another developer should add any lines to the loop further down the line.

_test/test_mpi/test_MESS_NAMES_factory.m
abuts: That's if you use git. As a first time look its better to have it obvious. and if it is a distributive, not related to git?

nickbattam-tessella: The `master` version of Horace+Herbert are now on GitHub. Be definition we *are* using Git.

herbert_core/classes/MPIFramework/@MessagesFilebased/MessagesFilebased.m
abuts: I have changed it for complacency sake, but not sure if the change is for good. If it was absolutely clear what is going on before. It needs some brain processing power to understand what is happening now. And performance here is not an issue at all.

nickbattam-tessella: the one-liner is a bit cryptic (and possibly wrong). Using a bool comparison to set a bool variable is just ugly.

_test/test_mpi/test_MESS_NAMES_factory.m
nickbattam-tessella: Changing the version number management is already a separate issue ticket; so don't worry about this for now.

_test/test_mpi/test_CPP_MPI_exchange.m
nickbattam-tessella: I mean is this one test function testing one thing or testing many things?

herbert_core/classes/MPIFramework/messages/MESS_NAMES.m
abuts: on the verge of acceptance.... despite we are not limited by keyword length, an abbreviation name2code_map reads (IMHO) much easier then the name_to_code_map....

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
abuts: done

_test/test_mpi/test_CPP_MPI_exchange.m
abuts: it is testing send-receive operation or rather correct matlab send-receive interface as nothing is actually send/received here.

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
abuts: I think, wrapping test code in a test code would be overkill in this situation

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
```diff
@@ -1,25 +1,102 @@
...
-		this->close();
-	}
+    //----------------------------------------------------------------------------------
+    // The methods used in unit tests -- have no meaning in real communications
 
-	int labIndex;  // index of the current MPI lab (worker)
-	int numProcs; // total  number of MPI labs (workers)
+    static bool MPI_wrapper_gtested;
+    // get access to the asynchroneous messages queue
+    std::list<SendMessHolder>* get_async_queue() {
```
hsautessella: Is there a reason we're returning a poiner to the `asyncMessList` object here and not the object itself? i.e. 

```cpp
const inline std::list<SendMessHolder>& get_async_queue() {
    return this->asyncMessList
}
```
We should `const` wherever possible, and `inline` our getters (tells the compiler to simply put the returned object in place, skipping the function call).

Same thing for `get_sync_queue` below.

_LowLevelCode/CPP/test/cpp_communicator.tests/cpp_communicator_test.cpp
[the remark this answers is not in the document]
hsautessella: So `MessCache` is a `std::list<SendMessHolder>*` `auto &message` would return type `SendMessHolder` whilst `auto &&message` would return type `&SendMessHolder`, i.e. a reference to the object (not sure why you'd need this though)  https://en.cppreference.com/w/cpp/language/range-for

_LowLevelCode/CPP/cpp_communicator/MPI_wrapper.h
abuts: because this may cause object copying and duplication I want to change object in-place. 
Accepting reference would to that, but then the receiver should always be with & and I think its more complicated.

abuts: const pointer would probably be ok, but as its for tests, I wanted maximal flexibility (not used it but you never know in advance)

hsautessella: But the `&` in the function type definition means there will be no copying, you're giving access to the unerlying object https://stackoverflow.com/questions/20091046/what-should-a-c-getter-return. 

In any case, if we're going to use pointers we should be using smart-pointers or we're going to run into memory leaks https://en.cppreference.com/book/intro/smart_pointers.

abuts: Let's move discussion on code style into #36 where one bug in this code (do not understand why it was the bug but testing clearly've showed it) was fixed.

Pointers are generally less safe, and proven to be a problem while keeping an allocated memory, but work  well for referring to existing objects.

abuts: very interesing discussion, but so petty when so much other work is wating attention....

abuts: here I was thinking of using the same pointer to switch from synchronous to asynchronous containters. This is possible and should be used in tests only. 
access to the underlying object is what is intended....

abuts: As I've already said, and just tested. whatever blog is telling
`    auto queue = wrap.get_async_queue();
    queue->rbegin()->theRequest = (MPI_Request)1;
`
makes and modifies the copy of the internal object if
`    std::list<SendMessHolder>& get_async_queue() {
        return this->asyncMessList;
    }
`
you need `auto &` to modify object and this is ugly.

Issue #36, opened 2020-01-08
Serialize MException


abuts:
This is part of bigger Horace ticket https://github.com/pace-neutrons/Horace/issues/70 on asynchronous communications. These communications request messages serialization and custom serialization needs to be deployed. Unfortunately, Matlab MException class, used in all Matlab error can not be serialized by existing tools, so small addition to messages, asllowing to serialize this class is necessary for proper MPI framework operations.

hsautessella:
closed by #38

Issue #33, opened 2020-01-07
Run Matlab tests in test step of build script


hsautessella:
The build script `tools/build_config/build.sh` created in #30 does not currently run Matlab tests. It should be edited to do so.

In order for the tests to work with Jenkins Matlab will need to exit on a non-zero error code if any tests fail. This blog post may be useful https://blogs.mathworks.com/developer/2015/01/20/the-other-kind-of-continuous-integration/

abuts:
I do not think it should exit on non-0 code (unless you hope to achieve to run Matlab in Parallel which I would not advice as initial step. The blog refers to different unit test suite from one we use. I have not read the full blog yet, but our test suite generates reports and write results for all tests package. 


Try runtests "test_folder" 
runtests test_file
or validate_herbert 
first from Matlab to try the existing testing procedure. 
then you can include any level of this type of testing in the script

hsautessella:
The reason I said that it needed to exit on a non-zero code is that: how does Jenkins know that a test has failed?

The tests will probably be run from within the `build.sh` script, using a command like `matlab -nodisplay -nosplash -r "validate_herbert; exit"`. But as far as Jenkins is concerned, the Matlab process it spawns will start and exit normally and the build will pass. Exiting with a non-zero code tells Jenkins that something has gone wrong and that the build should fail.

Or am I missing something?

abuts:
you need to modify validate_horace/herbert script to accept -run_as_script option and if this option is provided 
use `quit` (see help `quit`) with code depending on the results of the matlab unit tests. 
then you do not need to add exit to the end of the execution script as described in the blog

PR #37, opened 2020-01-10
Run Matlab tests in build script

hsautessella:
This PR adds commands so that CTest will run Matlab tests alongside the C++ ones.

The `_test/CMakeLists.txt` contains calls to CMake's `matlab_add_unit_test`which, in our case, will spawn a Matlab session and run `runtests()` on the given directory. It also allows commands to be executed before running the tests, e.g. `herbert_init`.

The `herbert_on.m`, `worker_v2.m` and `worker_v2.m` templates are copied into the CMake build folder. These are required to run the Matlab mpi tests. The path they are copied into is added to the `MATLABPATH` environment variable when the tests are run, so all spawned Matlab workers can see them.

The `mslice_objects` and `mpi` tests are failing and have been disabled - for now. #39 is open to fix the MSlice test, another ticket should be opened if the MPI tests are still faiiling after #36 goes in.


_test/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+    "test_instrument_classes"
+    "test_docify"
+    "test_admin"
+    "test_mpi"
+)
+
+foreach(_test_dir ${TEST_DIRECTORIES})
+    matlab_add_unit_test(
+        NAME "Matlab.${_test_dir}"
+        CUSTOM_TEST_COMMAND
```
nickbattam-tessella: defining this inline is inelegant; a function with `_test_dir` as an argument would be neater

herbert_core/admin/@opt_config_manager/private/find_comp_type_.m
```diff
@@ -60,6 +61,12 @@
         n_profile=n_profile+1;
     end
     pc_type = types{n_profile};
+
+end
+
+is_jenkins_true = is_jenkins()
```
nickbattam-tessella: why are you evaluating `is_jenkins()` here and on the next line ... is this variable needed?

herbert_core/admin/is_jenkins.m
```diff
@@ -0,0 +1,16 @@
+function [is_jenkins_pc] = is_jenkins()
+% Attempt to determine if the PC we're running on is a Jenkins server.
+%   This is detected through environment variables that are always set by
+%   Jenkins.
```
nickbattam-tessella: if they are set (presumably by something else) do you need to check all three, unless you're using them?

herbert_core/admin/@opt_config_manager/private/find_comp_type_.m
hsautessella: Ah thanks, this was leftover from debugging

herbert_core/admin/is_jenkins.m
hsautessella: They're not really used no, I guess I was just erring on the side of caution and using 3 values that are all definitely set. Just in case a user accidentally had one set. I can prune it down if you'd like?

abuts:
@hsautessella  It will be nasty problem if you have worker in admin folder and the herbert is not initialized properly on a slave Matlab session. If the slave worker script fails on initialization, the worker just hands up and you have no idea what is going on (on Windows you can look at slave session but that's it)
This is why worker script (e.g. worker_v2) is normally placed together with herbert/horace initalization (horace_on, herbert_on). You can initalize master node package by different means, but if herbert/horace_on are not available for workers, you are in trouble. 

This is why the scipts in admin folder are templates and copied as m files ontly to initialization place

hsautessella:
Accidentally pushed with Alex's dev branch merged in. I've reverted the merge with 9e59fd2, so this should only contain my work

herbert_core/admin/is_jenkins.m
nickbattam-tessella: it's fine

nickbattam-tessella:
#33

Issue #42, opened 2020-01-17
herbert's opt_config_manager fails if users put Horace_init on the path


abuts:
opt_config_manager helps users to configure their machines. Its oddity is that it is located in Herbert but need to know about Horace configurations. 

If user, sets up Horace as described in [method 1](http://horace.isis.rl.ac.uk/Download_and_setup#Standard_Installation), horace fails. This is a bug which should be fixed.

hsautessella:
#closed by #43

PR #43, opened 2020-01-17
Fixing opt_config_bag

abuts:
this is the fix for ticket #42. The fix is simple and contain all description of what went wrong withtin the class file itself.


herbert_core/admin/@opt_config_manager/opt_config_manager.m
```diff
@@ -159,7 +183,7 @@ function save_configurations(obj,varargin)
             % and return pc type.
             %
             % A pc type is a string, describing the computer from point of
-            % view of using it for high performance communications.
+            % view of using it for Horacing.
```
nickbattam-tessella: `Horacing`?

abuts: This term was used unofficially during Horace training courses. It sounds fisy so I think, its fun to have it in docs :)

PR #44, opened 2020-01-20
Herbert packaging

hsautessella:
This PR writes the necessary CMake scripts to package Herbert for distribution.

This should also make Herbert `make install`able.

To create a package, you should just have to run `cpack -G <generator>` in the build directory. Or run `./tools/build_config/build.sh --package`. A list of generators is available here https://cmake.org/cmake/help/v3.13/manual/cpack-generators.7.html

https://github.com/pace-neutrons/Horace/issues/35


tools/build_config/Jenkinsfile
```diff
@@ -91,6 +99,8 @@ pipeline {
                     )
                 ]
             )
+
```
nickbattam-tessella: unnecessary whitespace

admin/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# Function to install a file whose name ends in .template
+function(install_template_file)
```
nickbattam-tessella: does this function behave correctly if you call it with a non-template file name?

CMakeLists.txt
```diff
@@ -13,21 +13,38 @@ option(BUILD_FORTRAN "Build the Fortran Mex libraries" OFF)
...
+
+# =============================================================================
+# Install commands
+# =============================================================================
+include(herbert_CPackConfig)
+install(
+    DIRECTORY "herbert_core/"
+    DESTINATION "Herbert"
+    USE_SOURCE_PERMISSIONS
+    PATTERN "*.m~" EXCLUDE
```
nickbattam-tessella: are there C++ build files still kicking around that will get picked up (`.o`, `.obj`)?

cmake/herbert_FindGTest.cmake
```diff
@@ -59,6 +60,8 @@ if(RESULT)
     message(FATAL_ERROR "Build step for googletest failed: ${RESULT}")
 endif()
 
+# We don't want to install GTest libs when we run `make install`
+set(INSTALL_GTEST OFF CACHE BOOL "Enable installation of googletest." FORCE)
```
nickbattam-tessella: isn't this already covered by the 
```
if(${BUILD_TESTS})
   include(herbert_FindGTest)
   enable_testing()
endif()
```
block in the top-level CMake file?

admin/CMakeLists.txt
hsautessella: What does "correctly" mean here? If it means the file will still be copied without error if it does not have `.template` in the name, then yes. The `string(REPLACE ...)` will do nothing if a match is not found.

CMakeLists.txt
hsautessella: Not in `herbert_core` no. We assume an out-of-source build. I have neglected to put the temporary Matlab file extension for Windows in though - I'll add that

cmake/herbert_FindGTest.cmake
hsautessella: This flag disables the `install` target being created for GTest. If we did not have it, when we run `make install` CMake will copy all the GTest libraries/public headers into our package.

nickbattam-tessella:
Issue created to document the agreed naming convention in an ADR: https://github.com/pace-neutrons/Horace/issues/105

PR #45, opened 2020-01-20
Fix config error when unit tests are not available

hsautessella:
Fix typo that was throwing an error instead of a warning - `test_path` → `tests_path`. 

Running `init_herbert.m` using the new package was throwing an error since the package does not contain the `_test` directory. This was caused by a typo. 

Add qualifier so that the warning will only be shown if a user is attempting to initialize the tests and the directory does not exist. Previously the warning/error was shown every time the package was started

### DIFF
--- a/herbert_core/configuration/@herbert_config/private/process_unit_test_path.m
+++ b/herbert_core/configuration/@herbert_config/private/process_unit_test_path.m
@@ -10,11 +10,13 @@ if isempty(rootpath)
 end
 tests_path = fullfile(rootpath,'_test');
 if ~(exist(tests_path,'dir')==7)
-    warning('HERBERT_INIT:invalid_setup',...
-        'Can not set-up access to the unit tests as no unint tests at %s are available',...
-        test_path);
-    tests_path = '';
-    return;
+    if init
+        warning('HERBERT_INIT:invalid_setup',...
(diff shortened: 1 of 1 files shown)


herbert_core/configuration/@herbert_config/private/process_unit_test_path.m
```diff
@@ -10,11 +10,13 @@
 end
 tests_path = fullfile(rootpath,'_test');
 if ~(exist(tests_path,'dir')==7)
-    warning('HERBERT_INIT:invalid_setup',...
-        'Can not set-up access to the unit tests as no unint tests at %s are available',...
-        test_path);
-    tests_path = '';
-    return;
+    if init
```
nickbattam-tessella: these two `if`s can be combined as `if init && ~(exist(test_path, 'dir')==7)`. That will bypass the (more expensive) directory existence check if it's not needed.
 
Ideally the `exist(path, 'dir') == 7` function should be a directory_exists(path) helper for clarity; 7 is an unhelpful magic number (although this is fairly standard Matlab syntax).

abuts: I think return should be out of the if (init) block as nothing to do if tests folder is not available (Matlab do not add non-existing folders to the path so nothing to remove)

I've fixed it

PR #47, opened 2020-01-20
worker_v2 bug

abuts:
… DO_LOGGING variable is not available for worker_v2. Fixed

Trivial bug identified after close look at code of worker_v2

### DIFF
--- a/admin/worker_v2.m.template
+++ b/admin/worker_v2.m.template
@@ -146,42 +146,37 @@ while keep_worker_running
             % when its tested, workers are tested in single Matlab
             % session so it will hand up on synchronization
             if ~is_tested
-                if DO_LOGGING
-                    fprintf(fh,'Waiting at barrier for all do_job completion\n');
-                    disp('WORKER_4TESTS: Waiting at barrier for all do_job completion ')
-                end
-                % explicitly check for cancelation before data reduction
+                % explicitly check for cancellation before data reduction
                 is_canceled = je.is_job_canceled();
                 if is_canceled
                     error('JOB_EXECUTOR:canceled',...
-                        'Job canceled before synchonization after do_job')
+                        'Job canceled before synchronization after do_job')
                 end
                 
                 % when not tested, the synchronization is mandatory
                 je.labBarrier(false); % Wait until all workers finish their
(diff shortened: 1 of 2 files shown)


admin/worker_v2.m.template
```diff
@@ -146,42 +146,37 @@ while keep_worker_running
             % when its tested, workers are tested in single Matlab
             % session so it will hand up on synchronization
             if ~is_tested
-                if DO_LOGGING
-                    fprintf(fh,'Waiting at barrier for all do_job completion\n');
-                    disp('WORKER_4TESTS: Waiting at barrier for all do_job completion ')
-                end
-                % explicitly check for cancelation before data reduction
+                % explicitly check for cancellation before data reduction
                 is_canceled = je.is_job_canceled();
```
nickbattam-tessella: spelling: it should be `cancelled` in function and variable names

admin/worker_v2.m.template
```diff
@@ -146,42 +146,37 @@ while keep_worker_running
...
                 end
                 
                 % when not tested, the synchronization is mandatory
                 je.labBarrier(false); % Wait until all workers finish their
                 %                       job before reducing the data
             end
-            je.do_job_completed = true; % do not waitn at barrier if cancelation here
-            % explicitly check for cancelation before data reduction
-            %  the case of cancelation below
+            je.do_job_completed = true; % do not wait at barrier if cancellation here
```
nickbattam-tessella: what is the `do_job_completed` property you're setting here doing - the name is ambiguous? Executing a job-completed step or enabling a job complete function to be executed at some future point?

Every time it is set you have added comment about waiting (or not) at a barrier, which suggests that's what the method is controlling.

abuts: Yes -- name become umbiguous as initially it was only indicated job completeon. It changed its meaning after understanding that this is not the only barrier, the worker should bypath at throwing. 
Please, add this ticket to my backlog -- this is job closely related to description of the parallel framework, but I would not rather do it in a hurry between other jobs.

nickbattam-tessella: Created https://github.com/pace-neutrons/Herbert/issues/50

PR #49, opened 2020-01-20
Setup jenkins master pipeline

hsautessella:
The master pipeline of SL7 is set up on Jenkins. The changes in this PR are required for a successful build on the `Scientific-Linux-7-2018b` pipeline.

Currently, the pipeline is set up to build against this branch, this will be changed to `master` when this has been merged

### DIFF
--- a/tools/build_config/Jenkinsfile
+++ b/tools/build_config/Jenkinsfile
@@ -1,46 +1,22 @@
 #!groovy
 
-def getRepoURL() {
-  return sh(script: 'git config --get remote.origin.url', returnStdout: true)
-}
-
-def getCommitSha() {
-  return sh(script: 'git rev-parse HEAD', returnStdout: true)
-}
-
-def print_hello() {
-    return sh(script: 'echo hello', returnStdout: true)
-}
-
 def post_github_status(String state, String message) {
-    // Non-PR builds will not set PR_STATUSES_URL - in which case we do not
-    // want to post any statuses to Git
-    if (env.PR_COMMIT_SHA) {
-        commit_sha = env.PR_COMMIT_SHA
-    } else {
-        commit_sha = getCommitSha()
+    withCredentials([string(credentialsId: 'GitHub_API_Token',
+            variable: 'api_token')]) {
+        sh """
+            if [[ -z ${PR_STATUSES_URL} ]]; then \
+                commit_sha=$(git rev-parse HEAD); \
+                PR_STATUSES_URL="https://api.github.com/repos/pace-neutrons/Herbert/statuses/${commit_sha}"; \
+            fi && \
+            curl -H "Authorization: token ${api_token}" \
+                --request POST \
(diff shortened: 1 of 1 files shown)


tools/build_config/Jenkinsfile
```diff
@@ -1,18 +1,22 @@
...
-            curl -H "Authorization: token ${api_token}" \
-                --request POST \
-                --data '{"state": "${state}", \
-                         "description": "${message}", \
-                         "target_url": "$BUILD_URL", \
-                         "context": "$JOB_BASE_NAME"}' \
-                $PR_STATUSES_URL > /dev/null
-        """
-            }
+    // Non-PR builds will not set PR_STATUSES_URL - in which case we do not
```
nickbattam-tessella: Builds of `<named-branch>` (master or anything else) should still set build status flags if at all possible. 

If the build fails we need to know!

tools/build_config/Jenkinsfile
```diff
@@ -1,18 +1,40 @@
 #!groovy
 
+def getRepoURL() {
```
nickbattam-tessella: https://stackoverflow.com/questions/36547680/how-to-do-i-get-the-output-of-a-shell-command-executed-using-into-a-variable-fro suggests you can directly access the output of the shell command in Jenkins/Groovy without needing an intermediate file.:
```
return sh(script: 'git config --get remote.origin.url', returnStdout: true)
```
I you need the `trim()` you can probably append that.

tools/build_config/Jenkinsfile
```diff
@@ -1,18 +1,40 @@
...
-                         "description": "${message}", \
-                         "target_url": "$BUILD_URL", \
-                         "context": "$JOB_BASE_NAME"}' \
-                $PR_STATUSES_URL > /dev/null
-        """
-            }
+    // Non-PR builds will not set PR_STATUSES_URL - in which case we do not
+    // want to post any statuses to Git
+    if (env.PR_COMMIT_SHA) {
+        def repo_url = getRepoURL()
```
nickbattam-tessella: `repo_url` is the same on both branches of this `if`

tools/build_config/Jenkinsfile
```diff
@@ -1,18 +1,40 @@
...
+        def commit_sha = env.PR_COMMIT_SHA
+
+    } else {
+        def repo_url = getRepoURL()
+        def commit_sha = getCommitSha()
```
nickbattam-tessella: since `repo_url` is the same in both, can you replace this if with an inline if? 
```
def commit_sha = (env.PR_COMMIT_SHA) ?  env.PR_COMMIT_SHA : getCommitSha()
```

tools/build_config/Jenkinsfile
nickbattam-tessella: Created new ticket (#51) to resolve this issue; the simple extensions attempted (below) have failed.

PR #54, opened 2020-01-21
Update build version to match revised format

nickbattam-tessella:
Add Matlab release string
Remove Linux flavour

### DIFF
--- a/cmake/herbert_FindMatlab.cmake
+++ b/cmake/herbert_FindMatlab.cmake
@@ -17,6 +17,9 @@ the path to the directory containing Matlab's dynamically linked libraries
 ``Matlab_UT_LIBRARY``
 the path to the Matlab UT library
 
+``Matlab_VERSION``
+the Matlab release e.g. 2019b
+
 See the FindMatlab.cmake documentation for other variables defined by this
 module. You'll find the file bundled with your CMake installation.
 
@@ -24,6 +27,7 @@ module. You'll find the file bundled with your CMake installation.
 find_package(Matlab REQUIRED COMPONENTS MAIN_PROGRAM MEX_COMPILER)
 get_filename_component(Matlab_LIBRARY_DIR "${Matlab_MEX_LIBRARY}" DIRECTORY)
 get_filename_component(Matlab_BIN_DIR "${Matlab_MAIN_PROGRAM}" DIRECTORY)
+get_filename_component(Matlab_VERSION "${Matlab_ROOT_DIR}" NAME)
 
 # Find the libut library
 find_library(Matlab_UT_LIBRARY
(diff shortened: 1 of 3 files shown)


cmake/herbert_FindMatlab.cmake
```diff
@@ -17,13 +17,17 @@ the path to the directory containing Matlab's dynamically linked libraries
 ``Matlab_UT_LIBRARY``
 the path to the Matlab UT library
 
+``Matlab_VERSION``
+the Matlab release e.g. 2019b
```
hsautessella: This will have the `R` preceding the year.

```suggestion
the Matlab release e.g. R2019b
```

cmake/herbert_Version.cmake
```diff
@@ -1,38 +1,45 @@
+#[=======================================================================[.rst:
+herbert_Version
+-----------------
+
+Build a version release string:
+  <version>[-<date>]-<target>-<matlab>[-<sha>]
+
+ Optional elements are included based on the value of Herbert_RELEASE_TYPE
+ - Date included for "nightly" builds
+ - SHA included for non-"release" builds (i.e. "nightly" or "pull-request"
```
hsautessella: Missing a close bracket here

cmake/herbert_Version.cmake
```diff
@@ -1,38 +1,45 @@
...
+
+Variables required by the module
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``Herbert_RELEASE_TYPE
```
hsautessella: Maybe it would be good to mention here that if `Herbert_RELEASE_TYPE` is empty, not-defined, not `nightly` or not `release` then the default mode is `pull_request` - this is the format we'd want for dev builds as well (with the sha included)

cmake/herbert_Version.cmake
```diff
@@ -1,38 +1,45 @@
...
-    string(REPLACE " " "." ${ARGV1} ${${ARGV1}})
-    string(REPLACE "\"" "" ${ARGV1} ${${ARGV1}})
-endmacro()
-
 if(UNIX)
-    # Get distribution name from /etc/os-release
-    get_release_value("ID" _id)
-    get_release_value("VERSION_ID" _version_id)
-    set(Herbert_PLATFORM "${_id}.${_version_id}")
+    set(Herbert_PLATFORM "linux")
```
hsautessella: It might make sense to convert this block into 
```cmake
if(APPLE)
  ...
elseif(UNIX)
  ...
elseif(WIN32)
  ...
endif()

```

Note I think `APPLE` needs to come before `if(UNIX)`, since `if(UNIX)` can evaluate to true on some apple systems

cmake/herbert_Version.cmake
nickbattam-tessella: added `(default)`

cmake/herbert_Version.cmake
nickbattam-tessella: The APPLE branch will be added as part of https://github.com/pace-neutrons/Horace/issues/47

PR #56, opened 2020-01-23
Pacakge _test directory with Herbert

hsautessella:
The reason for this comes mostly from Jenkins+Horace. Horace pulls the
Herbert build package when it runs its tests, however the testing
framework is in Herbert! Hence the test framework needs to be present in
Herbert - at least for the time being.

Ideally we would not package the _test directory at all and Horace's
tests should all be written using the new Matlab test framework. This is
probably not possible, however we should find a way to minimise the size
of the Herbert package we ship to users, but still allowing Horace to
perform its full set of tests etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,13 +42,21 @@ add_subdirectory("admin")
 # Install commands
 # =============================================================================
 include(herbert_CPackConfig)
-install(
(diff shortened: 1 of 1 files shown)


hsautessella:
Merged for now as required for Horace builds. #57 should be completed to tidy this up

Issue #46, opened 2020-01-20
MPI tests hanging on Linux


hsautessella:
The MPI tests hang and time out on Linux. The tests are currently disabled for this reason. The output of this issue should be the tests being re-enabled and passing.

PR #58, opened 2020-01-23
Re-enable MPI tests and increase timout length

hsautessella:
It's possible the tests were not hanging, but just taking so long CTest was timing them out.

Hopefully an hour is long enough for the test to execute fully

refs #46


nickbattam-tessella:
Closing PR unmerged -- the ticket remains open

PR #59, opened 2020-01-24
Stop application_off removing paths twice

hsautessella:
The function `application_off` was calling `herbert_off` which removed all paths beneath Herebert's root - except for those in the MATLABPATH environment variable. `application_off` then removed the paths AGAIN, this time also removing the paths in MATLABPATH - which we do not want.

### DIFF
--- a/herbert_core/herbert_init.m
+++ b/herbert_core/herbert_init.m
@@ -128,10 +128,6 @@ for i=1:numel(application_init_old)
             disp(['Function ',app_name,'_off.m not found in ',rootpath])
             disp('Clearing rootpath and subdirectories from Matlab path in any case')
         end
-        paths = genpath(rootpath);
-        warn_state=warning('off','all');    % turn of warnings (so don't get errors if remove non-existent paths)
-        rmpath(paths);
-        warning(warn_state);    % return warnings to initial state
         cd(start_dir)           % return to starting directory
     catch ME
         cd(start_dir)           % return to starting directory
--- a/herbert_core/herbert_init.m
+++ b/herbert_core/herbert_init.m
@@ -130,7 +130,6 @@ for i=1:numel(application_init_old)
             paths = genpath(rootpath);
             warn_state=warning('off','all');    % turn of warnings (so don't get errors if remove non-existent paths)
             rmpath(paths);
-            addpath(getenv('MATLABPATH'));
             warning(warn_state);    % return warnings to initial state
         end
         cd(start_dir)           % return to starting directory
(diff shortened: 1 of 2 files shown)


nickbattam-tessella:
which paths are in MATLABPATH that aren't removed by `herbert_off`?

herbert_core/herbert_init.m
```diff
@@ -127,11 +127,11 @@ function application_off(app_name)
         else
             disp(['Function ',app_name,'_off.m not found in ',rootpath])
             disp('Clearing rootpath and subdirectories from Matlab path in any case')
+            paths = genpath(rootpath);
+            warn_state=warning('off','all');    % turn of warnings (so don't get errors if remove non-existent paths)
+            rmpath(paths);
+            warning(warn_state);    % return warnings to initial state
```
hsautessella: I think we should still have the logic preventing paths in `MATLABPATH` being removed

hsautessella:
@nickbattam-tessella 
> which paths are in MATLABPATH that aren't removed by `herbert_off`?

Generally `MATLABPATH` is empty, when Matlab starts it adds the user Matlab directory, usually `/home/user/Documents/matlab/` or `~\Documents\Matlab` on Windows.

When we build Hebert `herbert_on.m.template`, `worker4tests.m.template` and `worker_v2.m.template` are copied into the `build/herbert_config` directory (removing the `.template` from the name and updating the path in `herbert_on`). When tests are run through CTest, the `MATLABPATH` environment variable is set to point to `build/herbert_config` so when workers are spawned in MPI tests, they can see herbert_on and the worker files. 

Except, at the moment, when `herbert_init` is called before a test is run, it's removing `Herbert/build/herbert_config` from the path and not re-adding it.

herbert_core/herbert_init.m
abuts: in this form -- yes

abuts: I do not actually understand why it is removing the MATPABPATH if you have this variable outside of the package. 
You can add this logic to the removal code. 

but in the form I committed, the second removal would not occur so should be irrelevant to MATLABPATH issue

abuts:
you can add test on MATLABPATH settings to the removal script : -- just get its value within the code and add again after removal

abuts:
But as it is only a Jenkins issue, I would rather copy the worker/_on app outside of the Herbert folder.
Unless you are thinking about include setting up this variable into installation/intitialization scripts themselves

abuts:
@hsautessella  This will not work -- new addpath should be moved outside of try/catch block

hsautessella:
> This will not work -- new addpath should be moved outside of try/catch block

Why will it not work? Are we expecting `rmpath` to throw? It works for me

abuts:
If you run herbert_on twice, it will add MATLABPATH to the path first time, and will not do this for the second time
this may cause complex to identify bugs

abuts:
looks fine to me now

PR #60, opened 2020-01-24
Bug with worker_4tests_local

abuts:
Re #1 for some reason, worker_4tests_local has not been updated as it should be (equal to worker_v2) . Fixed


_test/test_mpi/worker_4tests_local.m
```diff
@@ -1,4 +1,4 @@
-function [ok,err_mess,je]=worker_4tests_local(worker_controls_string)
+function [ok,err_mess,je]=worker_v2(worker_controls_string)
```
nickbattam-tessella: does the file need renaming to match the changed function name?

abuts: no, it does not. Fixed

PR #62, opened 2020-01-24
Mpi tests separation

abuts:
The ticket separates the tests, which actually run communcations jobs from the tests which verify the code, involved in supporting the communication jobs. 

Separated to simplify the testing and observing what is actually failing on Jenkins


hsautessella:
I've added the `test_mpi_wrappers` tests to CTest. They're passing locally

hsautessella:
Rebuilding to see if #59 helps...

abuts:
Distributed computing toolbox is not available on the test node. Changing this test to not to run in such case.

PR #71, opened 2020-01-30
Reaname her_default_path to default_herbert_path

hsautessella:
This matches the convention used in horace_on.m.

Add indentation to smaller functions within the function script

### DIFF
--- a/admin/herbert_on.m.template
+++ b/admin/herbert_on.m.template
@@ -1,55 +1,51 @@
-function path=herbert_on(non_default_path)
+function path = herbert_on(non_default_path)
 % The function intended to switch Herbert on and
-% return the path were Herbert is resided or 
+% return the path were Herbert is resided or
 % empty string if Herbert has not been found
 %
 %
-% The function has to be present in Matlab search path 
+% The function has to be present in Matlab search path
 % and modified for each machine to know default Herbert location
 %
 % The default Herbert location is the place where Herbert_init script
 % can be found.
 %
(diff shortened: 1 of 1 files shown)


admin/herbert_on.m.template
```diff
@@ -1,55 +1,51 @@
...
-    start_herbert(her_default_path);
+    start_herbert(default_herbert_path);
 end
 path = fileparts(which('herbert_init.m'));
-%
-%
 
+% -----------------------------------------------------------------------------
 function start_herbert(path)
+    addpath(path);
```
nickbattam-tessella: the two `start_herbert` use-cases don't check that herbert_init.m exists

admin/herbert_on.m.template
```diff
@@ -1,55 +1,51 @@
...
-%
-%
-her_default_path='/usr/local/mprogs/Herbert';
-%
+
+default_herbert_path = '/usr/local/mprogs/Herbert';
+
 if exist('non_default_path','var') && (strcmpi(non_default_path,'where') || strcmpi(non_default_path,'which'))
-    path = find_her_default_path(her_default_path);   
+    path = find_herbert_path(default_herbert_path);
```
nickbattam-tessella: this code-path never calls `herbert_init`

hsautessella: It's not supposed to call `herbert_init`. It's just telling you where - if anywhere - Herbert is currently activated

Issue #29, opened 2019-12-23
Setup-CI


nickbattam-tessella:
This issue mirrors the equivalent Horace story https://github.com/pace-neutrons/Horace/issues/30

nickbattam-tessella:
Blocked as unable to create the webhooks in GitHub for the Herbert repostiory.

PR #74, opened 2020-01-31
Remove unused install-dir argument

nickbattam-tessella:
Remove unused argument that doesn't exist in the ps1 script


nickbattam-tessella:
Part of #29

PR #78, opened 2020-02-04
Set job names in Jenkinsfile

hsautessella:
Sets the job names in Jenkins PR and Branch builds.


nickbattam-tessella:
Should the Git-SHA be truncated to 6char (as used in the build artifact name)  (`.trim(6)`)? The full SHA doesn't fit on one line
Should the Branch name have a prefix (`Branch:`)  in the same way the `Git-SHA:` label is used?

PR #80, opened 2020-02-07
Extend CI build for multiple Matlab Versions - Windows

hsautessella:
This PR creates the necessary additions to the build scripts and Jenkinsfile for us to build against a specified version of Matlab in our CI and locally.

Our CMake now has the flag `-DMatlab_RELEASE=Rxxxx[ab]` for users to specify which version of Matlab they'd like to find. CMake will attempt to find this version and build against it. If it cannot find the version an error will be raised. This is most effective on Windows as Matlab is found from the registry, on Linux CMake will only find whichever Matlab is on your path.

Also created are the Jenkins pipelines to run the tests and builds using different versions of Matlab; at the moment only Windows with 2018b has been added. To create a pipeline using a different version of Matlab copy one of the existing pipelines on the platform you want to build on and change the default value of the `MATLAB_VERSION` parameter to whichever release you want to build against e.g. R2019b.

Documentation detailing the new build scripts args etc will be put into Horace when similar pipelines are set up there.

### DIFF
--- a/cmake/MatlabHelpers.cmake
+++ b/cmake/MatlabHelpers.cmake
@@ -80,3 +80,8 @@ macro(release_changed _output)
         set(${_output} TRUE)
     endif()
 endmacro()
+
+macro(set_matlab_release _value)
+    set(Matlab_RELEASE "${_value}" CACHE STRING
+        "The release of Matlab to find e.g. R2018b" FORCE)
+endmacro()
--- a/cmake/MatlabHelpers.cmake
+++ b/cmake/MatlabHelpers.cmake
@@ -1,0 +1,82 @@
+set(MATLAB_VERSIONS_MAPPING
+    "R2019b=9.7"
+    "R2019a=9.6"
+    "R2018b=9.5"
+    "R2018a=9.4"
+    "R2017b=9.3"
+    "R2017a=9.2"
+    "R2016b=9.1"
+    "R2016a=9.0"
+    "R2015b=8.6"
+    "R2015a=8.5"
+    "R2014b=8.4"
+    "R2014a=8.3"
+    "R2013b=8.2"
+    "R2013a=8.1"
+    "R2012b=8.0"
+    "R2012a=7.14"
+    "R2011b=7.13"
+    "R2011a=7.12"
+    "R2010b=7.11"
+
+    ${MATLAB_ADDITIONAL_VERSIONS}
+)
+
+#[=======================================================================[.rst:
+.. command:: matlab_get_version_from_release_name
+
+  Returns the version of Matlab (17.58) from a release name (R2017k)
+#]=======================================================================]
+macro(matlab_get_version_from_release_name release_name version_name)
+    string(REGEX MATCHALL "${release_name}=([0-9]+\\.?[0-9]*)"
+        _matched
+        ${MATLAB_VERSIONS_MAPPING}
+    )
+    set(${version_name} "")
+    if(NOT _matched STREQUAL "")
+        set(${version_name} ${CMAKE_MATCH_1})
+    else()
+        message(WARNING "[MATLAB] The release name ${release_name} is not registered")
+    endif()
+    unset(_matched)
+endmacro()
+
+#[=======================================================================[.rst:
+.. command:: matlab_get_release_at_path
+
+  Returns the release name (R2017k) of Matlab (17.58) given its root directory
+#]=======================================================================]
+macro(matlab_get_release_at_path _root_dir _release)
+    if(EXISTS "${_root_dir}/VersionInfo.xml")
+        file(READ "${_root_dir}/VersionInfo.xml" _version_info)
+        string(REGEX REPLACE
+            ".*<release>(R[0-9]+[ab])</release>.*" "\\1"
+            ${_release} "${_version_info}")
+    else()
+        execute_process(
+            COMMAND matlab "-help"
+            WORKING_DIRECTORY "${_root_dir}/bin"
+            OUTPUT_VARIABLE _help_contents)
+        string(REGEX REPLACE ".*Version: ([0-9]\\.[0-9]).*" "\\1"
+            _version "${_help_contents}")
+        string(REGEX REPLACE
+            ".*(R[0-9]+[ab])=${_version}.*" "\\1"
+            ${_release} "${MATLAB_VERSIONS_MAPPING}")
+    endif()
+endmacro()
+
+macro(root_dir_changed _output)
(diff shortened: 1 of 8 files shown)


cmake/herbert_FindMatlab.cmake
```diff
@@ -17,26 +17,63 @@ the path to the directory containing Matlab's dynamically linked libraries
...
 module. You'll find the file bundled with your CMake installation.
 
 #]=======================================================================]
-find_package(Matlab REQUIRED COMPONENTS MAIN_PROGRAM MEX_COMPILER)
+include(MatlabHelpers)
+
+root_dir_changed(_root_changed)
+release_changed(_release_changed)
+if(_root_changed AND NOT _release_changed)
+    # If Matlab_ROOT_DIR has been changed but not Matlab_RELEASE, discard the
```
nickbattam-tessella: "If Matlab_ROOT_DIR has been changed but not Matlab_RELEASE" is exactly the same as `(_root_changed AND NOT _release_changed)`.

The comment should be why you're discarding the `_RELEASE` in favour of the `_ROOT` (and similarly in the `elseif()`

cmake/herbert_FindMatlab.cmake
```diff
@@ -17,26 +17,63 @@ the path to the directory containing Matlab's dynamically linked libraries
...
+
+if(${Matlab_FOUND})
+    set_matlab_release("${_found_release}")
+else()
+    set_matlab_release("${_CACHED_Matlab_RELEASE}")
+endif()
+
+# Throw error if the Matlab found does not match the Matlab requested by Matlab_RELEASE
+if(NOT "${_found_release}" STREQUAL "${Matlab_RELEASE}"
+        AND NOT "${_INPUTTED_MATLAB_RELEASE}" STREQUAL "")
```
nickbattam-tessella: `_INPUTTED_MATLAB_RELEASE` is never set of cleared, so I'm not sure how this condition can ever be hit

cmake/herbert_FindMatlab.cmake
```diff
@@ -17,26 +17,63 @@ the path to the directory containing Matlab's dynamically linked libraries
...
+    matlab_get_version_from_release_name("${Matlab_RELEASE}" _version)
+    find_package(Matlab EXACT ${_version} COMPONENTS MAIN_PROGRAM MEX_COMPILER)
+endif()
+
+# Get the release of the Matlab that's been found
+matlab_get_release_at_path("${Matlab_ROOT_DIR}" _found_release)
+
+# Set local cached versions of variables so changes on next run can be tracked
+set(_CACHED_Matlab_RELEASE "${Matlab_RELEASE}" CACHE INTERNAL "")
+set(_CACHED_MATLAB_ROOT_DIR "${Matlab_ROOT_DIR}" CACHE INTERNAL "")
```
nickbattam-tessella: inconsistent capitalisation here `_CACHED_MATLAB_ROOT_DIR ` vs `_CACHED_Matlab_RELEASE `

cmake/herbert_Version.cmake
```diff
@@ -12,10 +12,10 @@ Build a version release string:
 Variables required by the module
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-``Herbert_RELEASE_TYPE
+``Herbert_RELEASE_TYPE``
 Release type: "nightly", "release" or "pull-request" (default)
 
-``Matlab_VERSION
+``Matlab_VERSION``
```
nickbattam-tessella: variable has been renamed `Matlab_RELEASE`

tools/build_config/Jenkinsfile
```diff
@@ -78,13 +78,15 @@ pipeline {
               module load gcc/\$GCC_VERSION &&
               module load openmpi/\$OPEN_MPI_VERSION &&
               ./tools/build_config/build.sh --print_versions --build \
-                --cmake_flags "-DHerbert_RELEASE_TYPE=$RELEASE_TYPE"
+                  --cmake_flags "-DHerbert_RELEASE_TYPE=$RELEASE_TYPE" \
+                  --matlab_release \$MATLAB_VERSION
```
nickbattam-tessella: having this as `_VERSION` is now confusing isn't it; maybe we need to change the Jenkins variable too for consistency

tools/build_config/build.sh
```diff
@@ -35,14 +35,16 @@ function run_configure() {
...
+  local matlab_release=$4
+  local cmake_flags="${5-}"  # Default value is empty string
 
   cmake_cmd="cmake ${HERBERT_ROOT}"
   cmake_cmd+=" -G \"${CMAKE_GENERATOR}\""
   cmake_cmd+=" -DMatlab_ROOT_DIR=${MATLAB_ROOT}"
   cmake_cmd+=" -DCMAKE_BUILD_TYPE=${build_config}"
   cmake_cmd+=" -DBUILD_TESTS=${build_tests}"
   cmake_cmd+=" -DBUILD_FORTRAN=${build_fortran}"
+  cmake_cmd+=" -DMatlab_RELEASE=${matlab_release}"
```
nickbattam-tessella: have you tested that this fails if `Matlab_RELEASE` is inconsistent with `MATLAB_ROOT`?

cmake/herbert_FindMatlab.cmake
nickbattam-tessella: it looks like there is a piece of logic that we were depending on earlier that has now gone

cmake/MatlabHelpers.cmake
```diff
@@ -0,0 +1,87 @@
...
+macro(root_dir_changed _output)
+    if("${_CACHED_MATLAB_ROOT_DIR}" STREQUAL "${Matlab_ROOT_DIR}")
+        set(${_output} FALSE)
+    else()
+        set(${_output} TRUE)
+    endif()
+endmacro()
+
+macro(release_changed _output)
+    if("${_CACHED_Matlab_RELEASE}" STREQUAL "${Matlab_RELEASE}")
```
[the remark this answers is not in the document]
hsautessella: No - `STREQUAL` is defined and implemented by `if`, so cannot be used outside an `if` statement.

tools/build_config/build.sh
hsautessella: It throws an error the first time - not the second. Fix upcoming

hsautessella:
Test failures seem to be due to IO race conditions on the Windows server now it's running multiple jobs.  Both nodes are attempting to write to the Jenkins user's temporary directory.

tools/build_config/Jenkinsfile
nickbattam-tessella: Created #94 for this issue

PR #84, opened 2020-02-11
Set TMP environment variable in Jenkinsfile

### LINKED ISSUE #82: IO race conditions on Windows Jenkins
Now that there are multiple workers on the Windows Jenkins server, we are getting errors due to IO race conditions. Some tests write to the user's temporary directory, this means that the two separate jobs are attempting to write to the same place causing IO errors.

hsautessella:
This environment variable is what is returned by Matlab's `tempdir` function. This avoids an issue where different jobs running on the same machine attempt to write to the same place at the same time.

Fixes #82


hsautessella:
Closing as alternative fix is easier to implement on clean branch

Issue #83, opened 2020-02-10
Bug in current gen_sqw


abuts:
Current GenSQW in Horace is failing. It looks like something wrong with MPI configuration. 
Its unclear if this is a Horace or Herbert problem, but one of the steps to find this bug is proper error reporting in Herbert. 
This is why this issue is in Herbert

PR #85, opened 2020-02-13
gen_sqw bug

abuts:
This is partial fix of the issue, reported at #83 as more changes have to be done in Horace (which may request relevant changes in Herbert)
Here I am enabling better error reporting from failing MPI job and fixing serialization on rundata(h) class (partially in Horace and some unit tests for the change are also in Horace) 

More changes may be necessary


hsautessella:
Failures on Windows in this PR look as if they're due to #82

abuts:
@hsautessella  the first failure was caused by incorrect line ending settings when Jenkins was pulling the git, the second -- not so sure -- I believe it was fixed and certainly not repeated.

_test/test_data_loaders/test_rundata.m
```diff
@@ -345,6 +347,33 @@ function test_serialization_crystal(this)
...
+        function test_serialize_on_file(obj)
+            %
+            run=rundata(f_name(obj,'MAP11014.nxspe'));
+            db = run.serialize();
+            runr = rundata.deserialize(db);
+            
+            assertEqual(run,runr);
+        end
+        %
+        function test_serialize_in_memory(obj)
```
nickbattam-tessella: what is this test testing? The previous test has shown `run === runr`, this seems to test that `run === runr` with some fields removed

herbert_core/classes/data_loaders/@asciipar_loader/asciipar_loader.m
```diff
@@ -187,6 +187,12 @@
                 return
             end
             [this.det_par_,this.n_detinpar_,this.par_file_name_] = check_det_par(value);
+            % HACK:
```
nickbattam-tessella: what problem is this `HACK` solving?

_test/test_data_loaders/test_rundata.m
abuts: it checks that serialization/deserialization work

herbert_core/classes/data_loaders/@asciipar_loader/asciipar_loader.m
abuts: the next commit rectifies this.

nickbattam-tessella:
SL7-2017b build failure is expected; that build configuration is not currently working

PR #90, opened 2020-03-03
MPI tests hanging on linux

### LINKED ISSUE #46: MPI tests hanging on Linux

hsautessella:
## Before merging this PR
**Ensure an issue is open to re-enable disabled tests:**
    - `test_job_dispatcher::test_job_with_logs_3workers` a2c3e21
    - `test_job_dispatcher::test_job_fail_restart` 8543b3a

-----------------------------------------------------------

Quite a few things went into this PR:

- References to Horace have been removed from the parallel worker scripts
    The MPI tests were originally hanging - this was because the worker script was calling `horace_on` - with no Horace available - so the workers threw an error and the main thread was left waiting. 

- `load module OpenMPI` has been removed from the Jenkinsfile
    OpenMPI is the wrong MPI library - we need MPICH

- `validate_herbert` is now called to run tests in CTest
    The script restores config after it's been run and the command to run the test written in CMake has been significantly shortened

- MPICH 3.3a2 has been packaged within the repo _for Linux only_
    Nothing fancy done here. Directory `herbert_core/external` has been created and MPICH has been put in it. We may wish to create a separate repo to store our binary dependencies in (https://github.com/pace-neutrons/Horace/issues/138), but this PR is getting very big as it is and it's more important to get the tests running.
    I've put it in `herbert_core` since that's what's currently packaged _by Horace_. We can review the packaging structure in another PR.
    A separate PR can package MSMPI for Windows - but the system MPI on the build server seems to work for now
    The `ClusterMPI` class will pick up the `mpiexec` from the bundled MPICH

- The tests running more than parallel 2 workers have been disabled
    ANVIL's sl7 machines only offer us two cores. I've contacted them to try to resolve this, but in the meantime it's more important to get the other tests running.

Fixes #46


_test/CMakeLists.txt
```diff
@@ -29,19 +29,9 @@ foreach(_test_dir ${TEST_DIRECTORIES})
...
-                exit(1);\
-            end;\
-            exit(~res)"
-        ADDITIONAL_PATH
-            "${CMAKE_SOURCE_DIR}/admin"
-            "${CMAKE_SOURCE_DIR}/herbert_core"
-        TIMEOUT 300  # 5 mins
+            "validate_herbert ${_test_dir} -exit_on_completion -talkative"
+        ADDITIONAL_PATH "${CMAKE_SOURCE_DIR}/admin"
+        TIMEOUT 6200  # 2 hours
```
hsautessella: This probably doesn't need to be this long

hsautessella:
Issue  #92 open to re-enable tests

PR #96, opened 2020-03-06
Make CMake helper scripts useable by Horace

### LINKED ISSUE #95: Make CMake helpers general and include in build package
There's a lot of common CMake code between Herbert and Horace, and Herbert has come a bit further along than Horace. Instead of mirroring all the changes to the CMake files in Horace, we should make the Herbert CMake files general and package them in our nightlies. This will allow Horace to use the same ones, so we no longer have to maintain two sets of the same files!

Note that our `nightly-deploy` pipeline will strip Herbert of any CMake files before the package is pushed anywhere user-facing.

hsautessella:
There is a lot of duplication in the CMake scripts between Herbert and Horace. Recently, there have been quite a few updates to Herbert's CMake scripts that need mirroring in Horace. Instead of duplicating even more code between the repos, this PR makes Herbert's scripts more general, allowing them to be used by Horace. 

CMake scripts that are Herbert specific are kept in the top-level of the `cmake` directory and are prefixed with `Herbert_`. Scripts that are intended to be used by both libraries are kept within `cmake/shared` and are prefixed with `PACE_`.

This does require us to package the CMake scripts when we create a package for Horace to build against. We can strip these out before we distribute to users though.

Fixes #95


CMakeLists.txt
```diff
@@ -15,16 +15,17 @@ option(BUILD_FORTRAN "Build the Fortran Mex libraries" OFF)
 list(APPEND CMAKE_MODULE_PATH
     "${CMAKE_SOURCE_DIR}/cmake"
     "${CMAKE_SOURCE_DIR}/cmake/external"
+    "${CMAKE_SOURCE_DIR}/cmake/shared"
 )
 
 # Look for packages early so we can exit if they're not found
 include(herbert_FindMPI)
```
nickbattam-tessella: why is herbert_FindMPI still project specific? You could remove the "herbert_core" dependence with an additional argument

nickbattam-tessella: Is there any reason to leave `herbert_FindMPI` unchanged? The only `herbert` specific piece is `herbert_root` in a file path; that could become an argument to make it reusable

hsautessella: I did think about this, however we're not currently building anything against MPI in Horace, so thought it wasn't necessary at this point. It shouldn't be too difficult to make more general though, if you think there's some value in that

cmake/.cmake-find
```diff
@@ -0,0 +1,4 @@
+This is an intentionally empty file that can be used by other CMake projects to
```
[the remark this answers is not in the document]
hsautessella: It doesn't really make any difference - the file is never going to be read by any code, it's just there so the CMake in Horace can reliably identify the `cmake` directory

PR #102, opened 2020-03-19
Re-enable disabled job dispatcher tests

### LINKED ISSUE #92: Re-enable disabled job_dispatcher tests
- `test_job_dispatcher::test_job_with_logs_3workers` https://github.com/pace-neutrons/Herbert/commit/a2c3e2140179c97342a6d9160bd9697b7ab3a677
- `test_job_dispatcher::test_job_fail_restart` https://github.com/pace-neutrons/Herbert/commit/8543b3a32d24308a5e3c144fb59426a2eb0b772c

The tests disabled require 3 workers, but ANVIL's sl7 machines only offer us two cores. I've contacted them to try to resolve this. There may be some Matlab parallel toolbox configuration we can set to get the tests working, as it only seems to be the `parpool` versions of the tests that fail. 

I've been informed that the sl7 machines should have 2 physicals cores, with hyper-threading. So they should make 4 workers available to Matlab.

```
===== Test Case Failure =====
Location: 
Name:     test_job_dispatcher_parpool::test_job_with_logs_3workers

job test_parpool_3workers_YNNFMPQJJY requested 3 workers while the cluster allows only 2

===== Test Case Failure =====
Location: 
Name:     test_job_dispatcher_parpool::test_job_fail_restart

job test_job_parpool_fail_restart_OGPRFCHHKM requested 3 workers while the cluster allows only 2

```

hsautessella:
By default, Matlab's parpool cluster's have a size equal to the number of physical cores on the system. This was causing issues on the build server when running tests requiring 3 workers on a VM with 2 physical (and 4 logical) cores. This may also confuse users who may expect to be able to utilise logical cores.

I've added some logic to `ClusterParpoolWrapper` that will check the number of logical cores available, and compare it to the `n_workers` argument passed into `ClusterParpoolWrapper.init`. If more workers are requested than number of physical cores, but enough logical cores are available, the number of cores in the cluster is increased.

Fixes #92 
Fixes https://github.com/pace-neutrons/Horace/issues/159


herbert_core/classes/MPIFramework/get_num_cores.m
```diff
@@ -0,0 +1,16 @@
...
+%
+% Output:
+% -------
+%   physical      The number of physical cores on this PC
+%   logical       The number of logical cores on this PC
+%
+core_info = evalc('feature(''numcores'')');
+match_str = 'MATLAB detected: ([0-9]+) %s cores';
+
+[match, ~] = regexp(core_info, sprintf(match_str, 'physical'), 'tokens', 'match');
```
nickbattam-tessella: should this line processing be another helper function? you do the same line processing twice:
```
core_info = evalc('feature(''numcores'')');
physical = get_element(core_info, 'physical')
logical = get_element(core_info, 'logical')
```

nickbattam-tessella: if you're feeling enthusiastic the `get_element()` method can then be tested with some hard-coded test strings to verify it's pulling the correct data.

herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
```diff
@@ -114,6 +114,18 @@
             
             cl  = parcluster();
             
+            % By default Matlab only utilises physical cores; enable use of
+            % logical cores if required
+            if cl.NumWorkers < obj.n_workers
```
nickbattam-tessella: It's not obvious what this block of code is doing. `obj_n.workers` is the requested number, `cl.NumWorkers` is the number that will be created (default `num_physical_cores` if I understand correctly)

herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
```diff
@@ -114,6 +114,18 @@
...
+                [n_physical_cores, n_logical_cores] = get_num_cores();
+                if obj.n_workers <= n_logical_cores
```
nickbattam-tessella: does there need to be a `n_workers > n_logical_cores` else block that displays an (friendly) error, rather than waiting for the inevitable exception

herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
```diff
@@ -114,6 +114,18 @@
...
+                    cl.NumWorkers = obj.n_workers;
+                    if cl.NumWorkers > n_physical_cores
```
nickbattam-tessella: should these two steps be the other way around? 
```
if (obj.n_workers > n_physical_cores) 
    disp("...enabling hyperthreading...")
end
cl.NumWorkers = obj_n_workers
```

_test/test_mpi_wrappers/test_cluster_wrapper.m
```diff
@@ -158,6 +158,18 @@ function test_check_progress_disp_results(obj)
             assertEqual(cluster.log_value,ref_string);
             
         end
+        
+        function test_parpool_uses_logical_cores_if_requested(obj)
```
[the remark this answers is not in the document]
hsautessella: If a `ClusterParpoolWrapper` is instantiated with more workers than a system has physcal cores, then use of logical cores has been requested. Maybe `test_parpool_can_utilise_logical_cores` is clearer?

_test/test_mpi_wrappers/test_cluster_wrapper.m
```diff
@@ -158,6 +158,18 @@ function test_check_progress_disp_results(obj)
...
+            [physical_cores, logical_cores] = get_num_cores();
+            if physical_cores >= logical_cores
```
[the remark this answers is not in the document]
hsautessella: If `logical_cores == physical_cores` then the ` clust.init(physical_cores + 1, ...)` call below will fail because not enough cores will be available.

The `>` case won't happen, so that could be taken out. I think I just put it in because the rest of the test is not valid for that case, although it's probably better we just let the test fail, because something would obviously be very wrong.

_test/test_mpi_wrappers/test_cluster_wrapper.m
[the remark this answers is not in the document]
nickbattam-tessella: `test_parpool_utilises_logical_cores_when_n_gt_physical_cores` to go with `test_parpool_utilises_physical_cores_when_n_eq_num_physical_cores` and `test_parpool_utilises_physical_cores_when_n_lt_num_physical_cores`

_test/test_mpi_wrappers/test_cluster_wrapper.m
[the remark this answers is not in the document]
nickbattam-tessella: I would take the `>` out for clarity.

herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
hsautessella: The block below does this

herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
hsautessella: Really, the print statement isn't necessary, I put it in there as a debugging statement originally. I've removed it for now, but it can be put back in

herbert_core/classes/MPIFramework/@iMessagesFramework/iMessagesFramework.m
```diff
@@ -45,6 +45,8 @@
             if nargin>0
                 obj.job_id = varargin{1};
             else
+                % use process ID to seed the RNG so processes don't conflict
+                rng(feature('getpid'));
```
nickbattam-tessella: this obj_id generation should be extracted to a util function, the same non-trivial code is now in two places

nickbattam-tessella: we shouldn't reseed the random number generator before each call. seeds should be set once on each MATLAB process. If you're just wanting to generate a unique id for this job, concatenating PiD + datetime would give you something easily convertible to a string

hsautessella: Yeah, I'm planning on finding the right place for this tomorrow. This push was a test to see if this fixes things

abuts:
have some minor comments but ok, will make them in my branch

PR #103, opened 2020-03-24
Refactored data loaders

abuts:
This removes all unused code from data loaders and refactors data loaders for simplicity, reliability and improved custom serialization. 

Hopefully, it will fix number of MPI issues.


herbert_core/classes/data_loaders/@rundata/rundata.m
```diff
@@ -372,38 +382,41 @@
...
                     end
                     this.loader__.par_file_name = val;
                 end
             else
+                classname = class(this);
+                out = feval(classname);%(this,'data_file_name',data_fname,'par_file_name',val));
                 this = out.initialize(data_fname,val,this);
             end
-            %this = this.select_loader('data_file_name',data_fname,'par_file_name',val);
+            %this = this.select_loader_('data_file_name',data_fname,'par_file_name',val);
```
nickbattam-tessella: commented out code

herbert_core/classes/data_loaders/saveNXSPE.m
```diff
@@ -54,22 +54,22 @@ function  saveNXSPE(nxspe_filename,data2save,varargin)
...
+tf = struct();
 tf.S=data2save.S';
 tf.ERR=data2save.ERR';
 tf.en = data2save.en;
 tf.det_par = [ones(numel(data2save.det_theta),1),data2save.det_theta*(180/pi),...
     data2save.det_psi*(180/pi),data2save.det_dtheta,...
     data2save.det_dpsi,data2save.det_group]';
-[~,fn] = fileparts(nxspe_filename);
-tf.save(fn);
+%[~,fn] = fileparts(nxspe_filename);
```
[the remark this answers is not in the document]
abuts: Its internal agreement to mark protected and private properties with single underscore. This agreement was not always there and initital intent was to mark them like in python __ private and _ -- protected (You can not put in in front, as would be preferable) The __ left from the time, when we though that __ should indicate private.  Probably we should make a separate ticket and teplace all __ to _ -- __ found to be excessive. If you think it should be done, please assigne to me a separate maintenance ticket. 

Similar agreement seems reasonable to use to indicate private functions (in folder private, not a protected functions) Not strictly adhered and not implemented everywhere, but seems substantiallty improves the code readability. Should be discussed separately and modified on agreement -- much more work to do.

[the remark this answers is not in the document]
abuts: I will investigate windows failures in a moment.

herbert_core/classes/data_loaders/saveNXSPE.m
```diff
@@ -54,22 +54,22 @@ function  saveNXSPE(nxspe_filename,data2save,varargin)
 function rd = parce_fromwindow(data2save,nxspe_filename)
 % the routine takes fromwindow mslice structure and converts it into
 % rundata to save in nxspe file
-tf = memfile();
+tf = struct();
```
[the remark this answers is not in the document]
abuts: Is it there? I believe, I have removed it. Git keeps local folders on remote deletion but the folder should be empty

herbert_core/classes/data_loaders/@rundata/private/set_up_from_struct_.m
```diff
@@ -0,0 +1,21 @@
+function obj = set_up_from_struct_(struc)
```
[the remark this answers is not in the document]
abuts: yes -- these are three tests I've added on last commit. I hope they test everything in there.  If wrong, may add something.

herbert_core/classes/data_loaders/@rundata/rundata.m
abuts: will fix.

_test/test_mpi/test_job_dispatcher_herbert.m
```diff
@@ -13,6 +13,21 @@
             end
             this = this@job_dispatcher_common_tests(name,'herbert');
         end        
+        function test_job_fail_restart(obj, varargin)
+            if is_jenkins
+                return
```
hsautessella: Do you think we should disable these tests outright rather than just not running them on Jenkins (e.g. prepend the test with `DISABLED_`? It feels cleaner and makes it obvious there's a problem. Likewise for running tests on Linux/Windows.

abuts: I have enabled it this way, as my changes may affect this part of the test.

abuts: And I want to know, why and where it running and where is not. 
As this runs on windows and linux on my machine but have problem on Jenkins, we know where to dig.
I think it should be carteful

abuts: Additinally, there is separate problem on Windows. It mentioned in "disabled" text and should be addressed separately.

_test/test_data_loaders/test_asciipar_loader.m
```diff
@@ -81,7 +81,7 @@ function test_load_ASCII_par_binary(this)
             assertTrue(all(ismember({'filename','filepath','x2','phi','azim','width','height','group'},fields(par))));
             assertTrue(all(ismember(fields(par),{'filename','filepath','x2','phi','azim','width','height','group'})));
             assertEqual(28160,numel(par.x2))
-            assertEqual(28160,al.n_detectors)
+            assertEqual(28160,al.n_det_in_par)
```
nickbattam-tessella: where does the magic number `28160` come from? it appears in lots of the tests

_test/test_mpi/test_ParpoolMPI_Framework.m
```diff
@@ -15,24 +16,25 @@
...
             cl = parcluster();
             num_labs = cl.NumWorkers;
             if num_labs < 3
+                warning('Can not run test_send_receive_message, not enough workers');
                 return;
             end
             num_labs = 3*floor(num_labs/3);
-            if num_labs > 27
-                num_labs = 27;
+            if num_labs > 6
```
nickbattam-tessella: why has this dropped from `num_labs > 27` to `> 6`?

herbert_core/classes/data_loaders/@rundata/private/deserialize_.m
```diff
@@ -12,21 +12,4 @@
 end
 %size = size+8;
```
nickbattam-tessella: commented out `size` update

_test/test_data_loaders/test_asciipar_loader.m
abuts: Its the number of detectors and it sits in the data file, used for testing. there were 2 such numbers for two files, but it should go to one now, as the bigger one does not fit github -- ASCII is too big.

_test/test_mpi/test_ParpoolMPI_Framework.m
abuts: it may be a problem on Windows machine, when two jobs are running simultaneously. It could run 21 there, but is selecting 9, and two jobs may overload memory and processors. I do not remember the idea of the test, but 3 are certanly necessary to check all different types of communications and may be sufficent after careful thinking. I do not want to think about it so put 6

_test/test_data_loaders/test_asciipar_loader.m
nickbattam-tessella: If there are data files too large for GitHub they can go into the new network storage when we get that configured

nickbattam-tessella: Can that one number be extracted as a constant (`EXPECTED_NUM_DETECTORS`) or something since it's repeated so many time and matches up to a static value in the data file?

abuts: Not sure it improves readability, but as I am making changes anyway, will change this too.

nickbattam-tessella: Thanks Alex -- the advantage to me is that it explains what that number represents and reduces the likelihood of typos in typing the same 5-digit number lots of time

herbert_core/classes/data_loaders/@rundata/private/deserialize_.m
abuts: removed

_test/test_data_loaders/test_asciipar_loader.m
```diff
@@ -77,7 +77,7 @@ function test_load_ASCII_par_binary(obj)
             if ispc
                 assertEqual([fname,fext],obj.test_par_file);
             else
-                assertEqual([fname,fext],obj.test_par_file);
+                assertEqual([fname,fext],'demo_par.PAR');
```
nickbattam-tessella: is the test file `demo_par.PAR` or `demo_par.par`? You should be able to use the correctly-cased name on Linux and Windows.

herbert_core/classes/data_loaders/saveNxspe.m
```diff
@@ -54,22 +54,22 @@ function  saveNXSPE(nxspe_filename,data2save,varargin)
...
 tf.S=data2save.S';
 tf.ERR=data2save.ERR';
 tf.en = data2save.en;
 tf.det_par = [ones(numel(data2save.det_theta),1),data2save.det_theta*(180/pi),...
     data2save.det_psi*(180/pi),data2save.det_dtheta,...
     data2save.det_dpsi,data2save.det_group]';
-[~,fn] = fileparts(nxspe_filename);
-tf.save(fn);
+%[~,fn] = fileparts(nxspe_filename);
+%tf.save(fn);
```
hsautessella: Can these comments be removed?

admin/worker_4tests.m.template
```diff
@@ -21,6 +21,10 @@ err_mess = [];
 exit_at_the_end = true;
 if isempty(which('herbert_init.m'))
     herbert_on();
+    try
+        horace_on();
```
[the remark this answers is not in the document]
abuts: the same worker is used by Horace and if Horace is not on, it will fail.

herbert_core/classes/MPIFramework/@ClusterWrapper/ClusterWrapper.m
```diff
@@ -428,7 +428,8 @@ function check_availability(obj)
                     %                        end
                 else
                     failed = true;
-                    mess = fprintf('Java run error with ID: %d',term);
+                    mess = fprintf('Java run error with ID: %d\n',term);
+                    ok = false;
```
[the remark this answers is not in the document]
abuts: The Java process control is the thing I have not well mastered on different OS. The idea of this method is ok indicates Java process control working fine, and "failed" is related to state of the job. As you can notice from this piece of the code, its not very well thought through and improvement is needed. Unfortunately I do not know what improvement.

herbert_core/classes/data_loaders/saveNxspe.m
abuts: That's probably some debugging left.

admin/worker_4tests.m.template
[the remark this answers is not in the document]
nickbattam-tessella: that call can't be here; everything in Herbert should be able to be executed without the Horace code being installed.

_test/test_data_loaders/test_asciipar_loader.m
abuts: The code makes files on linux behave like the files on Unix. This was the user problem for the users who were taking random case files from Windows and complaining that they do not work on Linux. So, it is the test for picking mixed case files on any operating system

nickbattam-tessella: what is the purpose of this test:
i) that a Windows machine can read 'demo.par' and 'demo.PAR' 
ii) that a Linux machine can read files named 'demo.par' and 'demo.PAR' 

If that is the case we should have two tests: one reading `demo.par`, one reading `demo.PAR` executed on both OSs

admin/worker_v2.m.template
```diff
@@ -20,6 +20,10 @@ err_mess = [];
 exit_at_the_end = true;
 if isempty(which('herbert_init.m'))
     herbert_on();
+    try
+        horace_on();
```
nickbattam-tessella: same comments as for `horace_on` in worker_for_tests above. We cannot assume Horace is installed or available running Herbert code.

admin/worker_4tests.m.template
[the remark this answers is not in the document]
abuts: this is why its in try-catch block
if it is not installed, failure will be ignored. 
what other solution you can find?

admin/worker_v2.m.template
abuts: this is why its in try-catch block
if it is not installed, failure will be ignored. 
what other solution you can find?

PR #108, opened 2020-03-26
Manage version number for builds and releases

### LINKED ISSUE #104: Manage version number in Matlab and C++
**Outputs:**
The `herbert_version` function should return an up-to-date version string for Herbert. In dev builds and in the release package.

hsautessella:
This PR begins moving us away from using SVN versioned files. 

`herbert_version.m` now returns a version string of the form `<major>.<minor>.<patch>[.<git sha>]`, as do all the `mex` functions, where they used to return an SVN version string. The version string is generated by CMake at configure time. This means that the version is defined in one place: at the top of the top-level CMakeLists.txt file. If CMake has not been run, `herbert_version` will return `0.0.0.dev`

CMake will generate the `herbert_get_raw_version` which will return the current version number, `herbert_version` will supply an API to this. The file `version.h` will also be generated, this file will contain `#define VERSION <version>`, it can be included in any C++ file to gain access to the macro. Both these files have been added to `.gitignore`.

The scripts that update the SVN versions in files have been removed.

Fixes #104


_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
```diff
@@ -34,7 +35,7 @@ Inputs:  -- optional,
 Inputs:
   1  -- mode_name  -- the string 'labSend' identifies send mode
   2  -- pointer to the initialized MPI framework,
-  3  -- dest_id address (nuber) of the worker who should receive the message 
+  3  -- dest_id address (nuber) of the worker who should receive the message
```
nickbattam-tessella: typo: `numer` -> `number`
typo: `is_synchroneous` -> `is_synchronous`
typo: `assynchronously` -> `asynchronously`

admin/templates/version.h.template
```diff
@@ -0,0 +1,3 @@
+/* ${Herbert_AUTO_GENERATED_FILE_WARNING} */
+
+#define VERSION "${Herbert_SHORT_VERSION}"
```
nickbattam-tessella: should this be `VERSION` or `HERBERT_VERSION` ? Just thinking we could have namespace issues; it's quite a common word.

admin/validate_herbert.m
```diff
@@ -65,7 +65,7 @@
 initial_warn_state = warning();
 warning('off', 'MATLAB:class:DestructorError');
 % Generate full test paths to unit tests:
-rootpath = herbert_root();
+rootpath = fileparts(fileparts(which('herbert_init')));
```
nickbattam-tessella: why is this function changed?

cmake/shared/PACE_Version.cmake
```diff
@@ -54,3 +55,9 @@ if(NOT "${${PROJECT_NAME}_RELEASE_TYPE}" STREQUAL "release")
 endif()
 
 message(STATUS "${PROJECT_NAME}_FULL_VERSION: ${${PROJECT_NAME}_FULL_VERSION}")
+
+# Set the short version which is used in ${PROJECT_NAME}_version.[m|h] files
+set(${PROJECT_NAME}_SHORT_VERSION "${PROJECT_VERSION}")
```
nickbattam-tessella: this needs to be added to the header

cmake/shared/PACE_Version.cmake
```diff
@@ -27,8 +27,9 @@ formatted version string
 #]=======================================================================]
 
 set(${PROJECT_NAME}_FULL_VERSION "${PROJECT_VERSION}")
+string(TOUPPER "${${PROJECT_NAME}_RELEASE_TYPE}" ${PROJECT_NAME}_RELEASE_TYPE)
```
nickbattam-tessella: why the change from lower case to caps?

herbert_core/admin/herbert_version.m
```diff
@@ -1,64 +1,38 @@
...
-% if horace_version is called with parameter, the function
-% returns revision data (Matlab_SVN) as number rather then string
-% (convenient for versions comparison)
-%
-%
-% An pre-commit hook script provided as part of the package 
-% has to be enabled on svn and svn file properties 
-% (Keywords) Date and Revision should be set on this file 
-% to support valid Matlab versioning.
+% If 1 or fewer output arguments are specified, the full version string is
```
nickbattam-tessella: `If one or fewer .... If more than one ...`

admin/validate_herbert.m
hsautessella: See https://github.com/pace-neutrons/Herbert/pull/108/commits/cdd89474348a9e0d30fea8c93632fc7f555bda0d.

Tests were failing because of this . Maybe it needs its own issue/PR, but it's a very small change

nickbattam-tessella: Isn't the solution to move `herbert_root` somewhere that is always on the path? That function was created to remove the impenetrably complex `fileparts(fileparts(which(...)))` call

nickbattam-tessella: you might need to separate the `dps` piece out as that only does make sense in the test code.

nickbattam-tessella: Happy for that restructure to be a new PR since we've added code code duplication in "now working" code.

cmake/shared/PACE_Version.cmake
hsautessella: Just a usability thing, when you run CMake you can now do `cmake .. DHerbert_RELEASE_TYPE=RELEASE` or `cmake .. DHerbert_RELEASE_TYPE=release`

admin/templates/version.h.template
hsautessella: I've replaced this with a `constexpr` within a namespace. Using `constexpr` is better practice these days.

cmake/shared/PACE_Version.cmake
nickbattam-tessella: aren't you defining an input variable in that conversion? Would it be cleaner to store the TOUPPERed variable in a "local" `_release_type` ?

nickbattam-tessella: Aren't you redefining an input variable in that conversion? Would it be cleaner to store the TOUPPERed variable in a "local" scoped variable ?

hsautessella:
This needs a corresponding change in Herbert since the template files have now moved, the `FindHerbert.cmake` script needs to be updated.

VERSION
```diff
@@ -0,0 +1 @@
+3.3.0
```
nickbattam-tessella, line 1: Do we need to handle there being a newline at the end of the file in CMake and MATLAB?

herbert_core/admin/herbert_version.m
```diff
@@ -18,7 +18,13 @@
     if ~strcmp(ME.identifier, 'MATLAB:UndefinedFunction')
         rethrow(ME);
     end
-    VERSION = '0.0.0.dev';
+    try
```
nickbattam-tessella, line 21: should this inner try..catch block be a function?

VERSION
hsautessella: It won't build because CMake won't allow new-line characters in its version string. We might need to handle it on the Matlab side though

nickbattam-tessella: VERSION with a newline and trailing spaces will be stored in Git at some point, so we should be able to handle it if possible.

CMake has a STRIP command (`string(STRIP <string> <output variable>)`) which should let anything that is read be cleaned.

_LowLevelCode/CPP/cpp_communicator/cpp_communicator.cpp
```diff
@@ -93,7 +93,7 @@ Outputs: -- empty matrix.
 
 void mexFunction(int nlhs, mxArray* plhs[], int nrhs, const mxArray* prhs[])
 {
-    if (nrhs == 0 && nlhs == 1) {
+    if (nrhs == 0 && (nlhs == 0 || nlhs == 1)) {
```
nickbattam-tessella, line 96: what is `plhs[1]` if `nlhs == 1`?

nickbattam-tessella, line 96: if `nlhs==0` does `plhs[0]` exist, should we just return in that case?

hsautessella: If `nlhs == 0` and we define `plhs[0]` then the version will be assigned to `ans`. Or, we could just return and nothing is assigned to `ans`, it's up to us

nickbattam-tessella: Ahh I see -- that seems like a reasonable behaviour then. I don't really have any strong feelings, just worried you were assigning something to a potentially empty array pointer.

In any case the same change needs to go into `get_ascii_file.cpp`

Issue #113, opened 2020-03-30
Fix errors in Windows Jenkins jobs run concurrently


hsautessella:
Windows Jenkins build jobs that are started at the same time are causing each other to fail. The tests that are failing are within `test_mpi`.

At first look, it seems to be config/shared messaging folders conflicting (again).

PR #114, opened 2020-03-30
Fix error in Windows Jenkins jobs run concurrently

hsautessella:
This PR makes `MPI_Tests_Common` (which most MPI tests inherit from) call `set_local_parallel_config()` on construction. It will restore the previous config on `tearDown`. 

`set_local_parallel_config` moves the config directory to `tempdir()/config`, CMake sets the environment variable `TMP` to point to within the build directory, so each Jenkins job should now have a unique place it stores its config. We do a similar thing for the message exchange folders. This should prevent the two jobs reading/writing files in the same location.

Fixes: #113
Fixes: https://github.com/pace-neutrons/Horace/issues/168

### DIFF
--- a/_test/common_functions/MPI_Test_Common.m
+++ b/_test/common_functions/MPI_Test_Common.m
@@ -21,6 +21,7 @@ classdef MPI_Test_Common < TestCase
     properties(Access=private)
         current_config_folder;
         parallel_config_;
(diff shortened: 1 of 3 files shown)


hsautessella:
I've run this 4 times now and have had no failures - so this is ready for a review

PR #117, opened 2020-03-30
Custom-serializable mpi messages

abuts:
This is the "Outdated" ticket, to build serializable MException and, more general messages with saveobj/loadobj methods, used for custom serialization will be used in C++ MPI.


_test/test_mpi_wrappers/test_mess_cache.m
```diff
@@ -60,10 +60,10 @@ function test_get_cache_messages(obj)
         function test_cache_operations(obj)
             mess_cache.instance('delete');
             
-            mess_list{1} = aMessage('log');
-            mess_list{2} = aMessage('log');
+            mess_list{1} = LogMessage();
+            mess_list{2} = LogMessage();
             mess_list{3} = aMessage('completed');
```
nickbattam-tessella: What is the use-case for the generic message base class?? In the MPI framework we will have a well documented set of messages (of specific sub-types) that are passed around in each situation, so there are no instances of a "general message" created

abuts: That's a philosophical question where no defined answer exist, but I will try to explain the choice in the document I am starting to write.  First of all, generic Message contains generic code and there is some.  Second, there are number of cases when a name is a sufficient definition for a message ant it was convenient to define just the name. It was actually all what is needed for filebased framework. As messages develop and may need additional features, you would inherit aMessage and write specific classes to describe it. That is where specific messages classes become necessary. All messages are subscribed to messages factory (MESS_NAMES) and verified against subscription at message creation.

nickbattam-tessella: I guess the question is whether the base class (`aMessage`) or a specific `Message` `GeneralMessage` child is the better thing to use -- that way everything is equal.

Would be easy enough to change afterwards, so happy to leave it as it for now until you've finished the documentation.

_test/test_mpi/test_job_dispatcher_mpiexec.m
```diff
@@ -23,11 +23,11 @@ function test_job_fail_restart(obj, varargin)
             end
         end
         function test_job_with_logs_3workers(obj, varargin)
-            if ispc % fail in sequence of tests on winwods
+            if ispc % fail in sequence of tests on windows
```
nickbattam-tessella: why?

abuts: this would be a ticket I will write when dealing with MPI framework. This puzzles me -- I can not imagine what the reason is behind this.

hsautessella: This test will be re-enabled as part of #92

abuts: on Unix yes -- it runs, but on windows, even locally it hangs up. Passed couple of random times but usually -- is reliable hang-up

_test/test_mpi_wrappers/test_mess_cache.m
abuts: Yes, I am not sure what the final desighn should be. It evolved historically from simple to more complicated stuff, so every step in the chain was well grounded by minimal free energy, but if the final state is now optimal -- the question for the review.

Issue #121, opened 2020-04-02
Create separate default configurations for Jenkins windows and Jenkins Linux


abuts:
Jenkins servers use their default configuration to run test jobs. Our linix and windwos machines are veru different but use the same default configurations. This should be fixed.

PR #122, opened 2020-04-02
Create different default Jenkins configurations for Unix and Windows

abuts:
Very simple ticket #121.

### DIFF
--- a/_test/test_admin/test_pc_spec_config.m
+++ b/_test/test_admin/test_pc_spec_config.m
@@ -95,7 +95,7 @@ classdef test_pc_spec_config< TestCase
             assertTrue(is);
         end
         %
-        function test_set_pc_type(obj)
+        function test_set_wrong_pc_name(obj)
             cm = opt_config_manager();
             try
                 cm.this_pc_type = 'rubbish';
@@ -103,39 +103,39 @@ classdef test_pc_spec_config< TestCase
             catch ME
                 assertEqual(ME.identifier,'OPT_CONFIG_MANAGER:invalid_argument')
             end
+        end
+        function test_set_wrong_pc_id(obj)
+            cm = opt_config_manager();
             try
                 cm.this_pc_type = -1;
                 assertTrue(false,'No exception was thrown on invalid configuration')
             catch ME
                 assertEqual(ME.identifier,'OPT_CONFIG_MANAGER:invalid_argument')
             end
-            
-            cm.this_pc_type = 1;
-            assertEqual(cm.this_pc_type,'win_small');
-            assertEqual(cm.pc_config_num,1);
-            
-            cm.this_pc_type = 7;
-            assertEqual(cm.this_pc_type,'idaaas_large');
-            assertEqual(cm.pc_config_num,7);
-            
-            cm.this_pc_type = 'a_mac';
-            assertEqual(cm.this_pc_type,'a_mac');
-            assertEqual(cm.pc_config_num,3);
-            
+        end
+        %
+        function test_set_pc_type_by_id(obj)
+            cm = opt_config_manager();
             kpc = cm.known_pc_types;
-            nconf = numel(kpc);
-            for i=1:nconf
+            ntypes = numel(kpc);
+            for i=1:ntypes
                 cm.this_pc_type = i;
                 assertEqual(cm.this_pc_type,kpc{i});
                 assertEqual(cm.pc_config_num,i);
             end
-            for i=1:nconf
+        end
+        function test_set_pc_id_by_name(obj)
+            cm = opt_config_manager();
+            kpc = cm.known_pc_types;
+            ntypes = numel(kpc);
+            for i=1:ntypes
                 cm.this_pc_type = kpc{i};
                 assertEqual(cm.this_pc_type,kpc{i});
                 assertEqual(cm.pc_config_num,i);
             end
             
         end
+        
     end
 end
 
--- a/_test/test_admin/test_pc_spec_config.m
+++ b/_test/test_admin/test_pc_spec_config.m
@@ -1,8 +1,7 @@
 classdef test_pc_spec_config< TestCase
+    % Testing default configuration manager, selecting
+    % configuration as function of a pc type
     %
-    % $Revision:: 833 ($Date:: 2019-10-24 20:46:09 +0100 (Thu, 24 Oct 2019) $)
-    %
-    
     properties
         working_dir;
     end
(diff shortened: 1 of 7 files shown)


herbert_core/admin/@opt_config_manager/private/find_comp_type_.m
```diff
@@ -24,14 +24,18 @@
             pc_type = types{1};%windows small
         end
     end
-
+    if is_jenkins()
```
nickbattam-tessella, line 27: this `if` should be the first check inside the `if ispc` block as the other type detection is unnecessary:
```
if is_jenkins()
    pc_type = types{8};  % 'jenkins_win'
else
    [~,sys] = memory();
    etc...
end
```

herbert_core/admin/@opt_config_manager/private/find_comp_type_.m
```diff
@@ -56,17 +60,18 @@
...
     end
-
+    
     if hpc_computer
         n_profile=n_profile+1;
     end
     pc_type = types{n_profile};
-
+    %
+    if is_jenkins()
```
nickbattam-tessella, line 69: the same short-circuit logic applies here; if it's a jenkins run all the other checks are unnecessary, so this should be the first clause.

_test/test_admin/test_pc_spec_config.m
```diff
@@ -85,8 +121,21 @@ function test_set_pc_type(obj)
...
+            kpc = cm.known_pc_types;
+            nconf = numel(kpc);
+            for i=1:nconf
+                cm.this_pc_type = i;
+                assertEqual(cm.this_pc_type,kpc{i});
+                assertEqual(cm.pc_config_num,i);
+            end
+            for i=1:nconf
+                cm.this_pc_type = kpc{i};
+                assertEqual(cm.this_pc_type,kpc{i});
```
nickbattam-tessella, line 134: this test is confirming the previous line has executed:
```
a = b
assertEqual(a, b)
```

_test/test_admin/test_pc_spec_config.m
```diff
@@ -85,8 +121,21 @@ function test_set_pc_type(obj)
...
             assertEqual(cm.pc_config_num,3);
+            
+            kpc = cm.known_pc_types;
+            nconf = numel(kpc);
+            for i=1:nconf
+                cm.this_pc_type = i;
+                assertEqual(cm.this_pc_type,kpc{i});
+                assertEqual(cm.pc_config_num,i);
+            end
+            for i=1:nconf
```
nickbattam-tessella, line 132: what is the purpose of the second for-loop here?

herbert_core/admin/@opt_config_manager/OptimalConfigInfo.xml
```diff
@@ -231,8 +231,41 @@
...
+         <worker>worker_v2</worker>
+         <parallel_framework>herbert</parallel_framework>
+         <cluster_config>local</cluster_config>
+         <shared_folder_on_local/>
+         <shared_folder_on_remote/>
+         <working_directory/>
+      </parallel_config>
+   </jenkins_win>
+   <jenkins_unix>
+      <info>Settings for running on Jenkins Unix ANWIL build servers</info>
```
nickbattam-tessella, line 268: typo `ANWIL`

herbert_core/admin/@opt_config_manager/private/load_configuration_.m
```diff
@@ -1,48 +1,27 @@
...
 %
-conf = [];
 config_file = fullfile(obj.config_info_folder,obj.config_filename);
 if ~(exist(config_file,'file') == 2)
     warning('No existing configuration file %s found. Current configuration left unchanged',...
         config_file)
     return;
 end
 config_data = xml_read(config_file);
+obj.all_known_configurations_ = config_data;
```
nickbattam-tessella, line 17: is there any need to have this data as both class (`obj.all_known_configurations_`) and local (`config_data`) variable?

herbert_core/admin/@opt_config_manager/private/set_pc_specific_config_.m
```diff
@@ -0,0 +1,9 @@
+function obj = set_pc_specific_config_(obj,config_data,current_pc)
```
nickbattam-tessella, line 1: there's no need to pass `config_data` as an argument because it's an attribute of the `opt_config_manager` object.

herbert_core/admin/@opt_config_manager/OptimalConfigInfo.xml
```diff
@@ -242,15 +275,15 @@
       </herbert_config>
       <hor_config>
          <mem_chunk_size>20000000</mem_chunk_size>
-         <threads>8</threads>
+         <threads>2</threads>
```
nickbattam-tessella, line 278: does @hsautessella fix to logical vs physical cores mean that it's possible to run the ANVIL Linux nodes with `>2` threads?

hsautessella: That fix is not merged https://github.com/pace-neutrons/Herbert/pull/102

herbert_core/admin/@opt_config_manager/private/find_comp_type_.m
abuts: this function also returns memory size and number of processors. This is why the check is last.

herbert_core/admin/@opt_config_manager/private/find_comp_type_.m
abuts: again -- not only checks but should deal with number of processors. It is not correct here, as was not nesessary, but Harry is doing something for this in other places, so this change for the future. I made appropriate comments to the function.

_test/test_admin/test_pc_spec_config.m
abuts: this line checks if setters worked correctly. There are no abbreviation logic added to the code yet , (like its done in config clasees) as I do not think, it is important enough, but if it does one day, the test is there.

_test/test_admin/test_pc_spec_config.m
abuts: you can set up configuration both by number and by name

herbert_core/admin/@opt_config_manager/OptimalConfigInfo.xml
abuts: fixed

herbert_core/admin/@opt_config_manager/private/set_pc_specific_config_.m
abuts: fixed

herbert_core/admin/@opt_config_manager/OptimalConfigInfo.xml
abuts: at the moment it Horace tests was failing on Linux as it requested 5-6 workers
no Matlab parpool or mpiexec for some reason can run this. I wander how this test would be able

abuts: I believe, tests for cores and threads are different one, so there are range of omp tests which would use threads, whilc MPI would use both. Let it be too.

_test/test_admin/test_pc_spec_config.m
nickbattam-tessella: These two test cases should be split into two test functions then. 

Looking at the full function there are five distinct test cases be checked so these should be split out:
- throws error if unknown pc type
- throws error in invalid index  (which needs to check -1 and nconf+1 edges)
- sets 'amac' --> why is this in the sets PC case when it's not part of the know_pc_types array?
- sets pc by array index
- sets pc by name

PR #124, opened 2020-04-06
jenkins config folder

abuts:
In addition to creating Jenkins-specific job folder, there are couple of small changes added, the main one -- meaninful job_id as users would prefer to see something reasonable instead of random number


_test/test_mpi/test_job_dispatcher_parpool.m
```diff
@@ -98,12 +98,15 @@ function create_test_file(name)
...
+        % DISABLED on Jenkins
         function test_job_fail_restart(obj, varargin)
-            warning('test_job_dispatcher_parpool:test_job_fail_restart disabled')
-            return;
-            % disp('test_job_dispatcher_parpool:test_job_fail_restart')
-            % test_job_fail_restart@job_dispatcher_common_tests(obj, varargin{:})
+             if is_jenkins
+                warning('test_job_dispatcher_parpool:test_job_fail_restart disabled')
+                return;
+            else
```
nickbattam-tessella, line 106: no need for an `else` after `if xxx; return`

herbert_core/classes/MPIFramework/@MessagesFilebased/private/init_framework_.m
```diff
@@ -23,8 +23,8 @@
     end
 elseif(is_string(framework_info))
     % use process ID and time as job ID. This prevents clashes between processes
-    obj.job_id_ = sprintf('%05i', feature('getpid'), ...
-                          round(datetime('now').Second*1e3));
+    obj.job_id_ = sprintf('%s_%08i',framework_info,...
```
nickbattam-tessella, line 26: this job id generation logic is duplicated in `iMessagesFramework` so should be changed to there to be consistent

herbert_core/configuration/@config_store/config_store.m
```diff
@@ -34,6 +34,12 @@
             if is_virtual
                 newStore.config_folder_name_ = ['mprogs_config_',type];
             end
+            [is_virtual,build_name] = is_jenkins();
+            if is_virtual
+                [~,build_name] = fileparts(build_name); % remove all possible folder path's of the build name
```
nickbattam-tessella, line 39: this should in `tmpdir` as we're already managing `TMP` environment variable to create job-unique temp folder locations.

abuts: This one defines folder name. There is function called after that, which defines the folder location. 
I do not want the logic of the program running on Jenkins to be different from the one, running locally.  If we want folder to be in tmp and bring this logic into config_manager, the `create_config_folder` function should be changed. 
Currently we setting up tmp_dir in test configuration. This is fine -- `create_config_folder` picks up this. (or should pick it up)

herbert_core/classes/MPIFramework/@MessagesFilebased/private/init_framework_.m
abuts: fixed in the next commit

_test/test_mpi/test_job_dispatcher_parpool.m
abuts: ok, though not essential. Fixed as part  of group

Issue #133, opened 2020-04-09
Modify test suite to report test start/stop and execution time on request


abuts:
Some unit tests run on Jenkins are not propertly reporting their start/stop state which becomes a problem if investigating are they are randomly hang-up or not. The modifications add a property, whcih forces test suite report its state.

PR #135, opened 2020-04-09
modified tests to generate additional printout on request

abuts:
see the #133

### DIFF
--- a/_test/shared/matlab_xunit/xunit/TestCase.m
+++ b/_test/shared/matlab_xunit/xunit/TestCase.m
@@ -28,11 +28,12 @@
 %   Copyright 2008-2010 The MathWorks, Inc.
 
 classdef TestCase < TestComponent
-    
+   
     properties
         MethodName
     end
-
+    
+    
     methods
         function self = TestCase(testMethod)
             %TestCase Constructor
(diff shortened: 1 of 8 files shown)


_test/shared/matlab_xunit/xunit/TestCase.m
```diff
@@ -69,12 +70,20 @@
...
                     f = self.MethodName;
+                    name2print  = func2str(self.MethodName);
                 else
                     error('TestCase:invalid_argument',...
                         'Unknown Method name type');
                 end
+                if self.print_running_tests
+                    tStart = tic;                    
+                    fprintf('**************************************************** \n');
+                    fprintf('************  starging test: %s\n',name2print);
```
nickbattam-tessella, line 84: Typo: `Starting test`

_test/shared/matlab_xunit_ISISextras/@TestCaseWithSave/TestCaseWithSave.m
```diff
@@ -233,6 +233,9 @@
         
         % List of any paths to remove after test suite is completed
         paths_to_remove_= {};
+        
+        % 
+        print_running_tests_ = false;
```
nickbattam-tessella, line 238: Should this be declared be in the base `TestCase` class where it's used rather than here? 

That or move the decoration print statements into an overriding implementation
```
function self = TestCase(testMethod)
  tic()
  print "start <methodname>"
 
  super.TestCase(testMethod)
  toc()
  print "done <methodname> in <time>"
```

abuts: no -- that's leftower. Removed. The decorator is defined in TestSuite and the property itself -- in  TestComponent. To avoid multiple constructors, TestSuite generate its own TestCase for every method.

_test/shared/matlab_xunit/xunit/TestCase.m
abuts: fixed

PR #140, opened 2020-04-16
Bug in runtests

abuts:
This should resolve the bug and let runtests report failure if no tests were run

### DIFF
--- a/_test/disabled_tests.md
+++ b/_test/disabled_tests.md
@@ -7,7 +7,7 @@
   - test_labprobe_nonmpi (no ticket)
 
 - test_job_dispatcher_parpool.m
-  - test_job_submittion (no ticket)
+  - test_job_submittion (Its not a disabled test, but a tester to debug job submission on a cluster)
   - test_job_fail_restart (no ticket)  Disabled on Jenkins
 
 - test_job_dispatcher_mpiexec.m
--- a/_test/shared/matlab_xunit/xunit/TestSuite.m
+++ b/_test/shared/matlab_xunit/xunit/TestSuite.m
@@ -70,7 +70,7 @@ classdef TestSuite < TestComponent
         end
         
         
-        function did_pass_out = run(self, monitor)
+        function [did_pass_out,num_tests_runed] = run(self, monitor,num_tests_runed)
             %run Execute test cases in test suite
             %   did_pass = suite.run() executes all test cases in the test
             %   suite, returning a logical value indicating whether or not all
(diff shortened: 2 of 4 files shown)


abuts:
@nickbattam-tessella  The way component was added (or rather identified) was changed in previous PR.

_test/disabled_tests.md
```diff
@@ -7,7 +7,7 @@
   - test_labprobe_nonmpi (no ticket)
 
 - test_job_dispatcher_parpool.m
-  - test_job_submittion (no ticket)
+  - test_job_submittion (Its not a disabled test, but a tester to debug job submission on a cluster)
```
hsautessella, line 10: If this isn't an actual test, we could probably remove it from here.

abuts: may be name can be changed
The code was exteamly useful for debugging jobs on cluster so I do not want to remember how to do it when time come to run jobs on SCARF

And it is a good place to find it

abuts: as the first thing you will be doing testing cluster, is check that the cluster config is correct

This is the test, which should be run on SCARF only. For everything else its substantial waste of time. 
When we have is_jenkins & is_scarf -- this test should run

_test/shared/matlab_xunit/xunit/TestSuite.m
```diff
@@ -79,14 +79,17 @@
             if nargin < 2
                 monitor = CommandWindowTestRunDisplay();
             end
+            if nargin<3
+                num_tests_runed = 0;
```
nickbattam-tessella, line 83: `num_tests_run`

_test/shared/matlab_xunit/xunit/TestSuite.m
```diff
@@ -79,14 +79,17 @@
...
+            end
             
             monitor.testComponentStarted(self);
             did_pass = true;
             
             self.setUp();
             
             for k = 1:numel(self.TestComponents)
-                this_component_passed = self.TestComponents{k}.run(monitor);
+                [this_component_passed,num_tests_runed] = self.TestComponents{k}.run(monitor,num_tests_runed);
```
nickbattam-tessella, line 92: how is `num_tests_runed` being updated by this method?

abuts: TestComponent, which is acutal executive, adds one to its input value. 
But if you ask how the tests are running in detals -- I do not fully understand it. Its pretty engenious code. This is why Matlab get it for its unit tests (initially it was this code, though Matlab had spoiled it later)

_test/shared/matlab_xunit/xunit/TestSuite.m
abuts: would not spend time on this, but fine, changed and merging without test

PR #142, opened 2020-04-16
Move seed rng func into test shared

hsautessella:
This function was originally put in `Herbert/_test/common_functions` so that Horace tests could make use of it. However, `common_functions` is not part of the Herbert package and so the function was not available to Horace tests on the build servers.


hsautessella:
I'm going to cancel all the builds, so they'll show as failed

PR #148, opened 2020-04-23
Remove Fortran rebin and integration

### LINKED ISSUE #144: Remove FORTRAN rebin and integration code
Horace equivalent issue: https://github.com/pace-neutrons/Horace/issues/193
Refs: https://github.com/pace-neutrons/Horace/issues/129

hsautessella:
This PR removes the remaining Fortran routines and replaces their use with the equivalent Matlab functions. 

I've removed the Fortran build steps as well.

Fixes #144 
Fixes https://github.com/pace-neutrons/Horace/issues/193


herbert_core/utilities/classes/@IX_data_1d/private/integrate_1d_points.m
```diff
@@ -1,41 +1,167 @@
-function [sout,eout] = integrate_1d_points (x, s, e, xout, use_mex, force_mex)
+function [sout,eout] = integrate_1d_points (x, s, e, xout)
 % Integrates point data along axis iax=1 of an IX_dataset_nd with dimensionality ndim=1.
 %
-%   >> [sout,eout] = integrate_1d_points (x, s, e, xout, use_mex, force_mex)
+%   >> [sout,eout] = integrate_1d_points_matlab (x, s, e, xout)
```
nickbattam-tessella, line 4: this should still be changed to `integrate_1d_points `

tools/build_config/build.sh
```diff
@@ -103,7 +101,6 @@ function main() {
         -X|--build_tests) build_tests="$2"; shift; shift ;;
         -C|--build_config) build_config="$2"; shift; shift ;;
         -O|--build_dir) build_dir="$(realpath "$2")"; shift; shift ;;
```
nickbattam-tessella, line 103: The `gfortran -- version` call should be deleted from `print_package_versions`

admin/herbert_mex.m
```diff
@@ -27,9 +27,9 @@ function herbert_mex(varargin)
 % root directory is assumed to be that in which mslice_init resides
 
 % list of keys the script accepts
-options={'-prompt','-setmex','-CPP','-FOR','-keep_lib','-use_lib','-missing'};
+options={'-prompt','-setmex','-CPP','-keep_lib','-use_lib','-missing'};
```
nickbattam-tessella, line 30: references to FORTRAN still in he function header

herbert_core/classes/data_loaders/@spe/private/put_spe.m
```diff
@@ -26,18 +26,16 @@
 filename=[name,ext];
 filepath=[path,filesep];
 
-% Prepare data for Fortran routine: must get rid of NaNs
+% Remove NaNs from data
```
nickbattam-tessella, line 29: was this block of code *only* required for FORTRAN? in which case the whole section can be dropped rather than the comment changing

herbert_core/classes/data_loaders/@spe/private/put_spe.m
```diff
@@ -26,18 +26,16 @@
...
     data.S(index)=null_data;
     data.ERR(index)=0;
 end
 
 % Prepare data for Matlab: - must ensure no data has exponents outside range -99 to +99
-% The problem is that although the fortran will write e.g. -1.234E-008 as -1.234-008, which
-% the fortran read routines will cope with, the Matlab and C++ reads will not cope. Matlab
-% sees -1.234-008 as two numbers, and C++ sees this as one number, but does not interpreet the
-% exponent.
+% The problem is that Matlab sees -1.234-008 as two numbers, and C++ sees this
```
nickbattam-tessella, line 37: this is a problem triggered by a FORTRAN write; which is no longer happening; again do we need this check and clean?

nickbattam-tessella:
`get_spe_matlab` and `put_spe_matlab` implementations of `get_spe` and `put_spe` still exist but these are the only implementations. Should these be merged in the same way the `integrate_*` implementations have been?

herbert_core/classes/data_loaders/@spe/private/put_spe.m
hsautessella: My thinking was that I didn't want to change the way we're writing files. I don't know if there are any other routines/workflows that count on the fact that these SPE files do not contain NaNs. I felt leaving these blocks was safer. We also retain backwards compatibility this way.

I think also, `.spe` is being phased out in favour of `.nxspe` anyway, so I don't know if it's worth spending the time working out if it's ok!

nickbattam-tessella: That's sensible - the only downside is that if we leave the code it will need to be tested.

abuts: .spe has been phased off, but people still have their old spe around. The code still need to undertstand, so reliability instead of performance is critical. Writing spe still may be requested too for compartibility with old programs and codes. It is so nasty problem -- you do not know what exactly old program users still want so its scary to change anything in this code.

herbert_core/classes/data_loaders/@spe/private/put_spe.m
abuts: in SPE it still need to write as -1E+30 or whatever, for compartibility with old crap may laying around, so the replacing NaNs block is necessary.

herbert_core/classes/data_loaders/@spe/private/put_spe.m
mducle: If we have an archive of old Horace released versions, we can just tell users who want to use an old SPE file that they can just use the old version of Horace... (otherwise deprecation doesn't mean anything!)

nickbattam-tessella: That's a good point; it's a massive overhead to try to keep backward compatibility for increasingly rarely used features.

We can make breaking changes so long as we update the major version number and notify our users that it's going to happen.

PR #149, opened 2020-04-23
Update find root nexus

abuts:
Minor change to make find_root_nexus report correct errors in case of incorrectly written user files

### DIFF
--- a/_test/test_data_loaders/test_loader_utilites.m
+++ b/_test/test_data_loaders/test_loader_utilites.m
@@ -44,25 +44,25 @@ classdef test_loader_utilites< TestCase
        function test_correct_DSinCorrectFolder_found(this)
             [DS_info,ds_path] = find_dataset_info(f_name(this,'MAP11014.nxspe'),'data','data'); 
             
-            assertEqual(DS_info.Dims,[30,28160]);
+            assertEqual(DS_info.Dataspace.Size,[30,28160]);
             assertEqual(ds_path,'/11014.spe/data/data');                              
        end           
        function test_correct_DSinCorrect2Folder_found(this)
             [DS_info,ds_path] = find_dataset_info(f_name(this,'MAP11014.nxspe'),'data','azimuthal'); 
             
-            assertEqual(DS_info.Dims,28160);
+            assertEqual(DS_info.Dataspace.Size,28160);
             assertEqual(ds_path,'/11014.spe/data/azimuthal');                              
        end     
       function test_correct_DS2_found(this)
             [DS_info,ds_path] = find_dataset_info(f_name(this,'MAP11014.nxspe'),'','azimuthal'); 
             
-            assertEqual(DS_info.Dims,28160);
+            assertEqual(DS_info.Dataspace.Size,28160);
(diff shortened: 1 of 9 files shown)


herbert_core/utilities/hdf_nexus/find_root_nexus_dir.m
```diff
@@ -61,26 +72,28 @@
...
 % Matlab version specific function to obtain correct NeXus version
-mat_ver_array=datevec(version('-date'));
-
-if mat_ver_array(1)<=2009
-    [fp,shortName] = fileparts(DS.Attributes.Name);                    
-    ver = hdf5read(hdf_fileName,[DS.Name,'/',shortName]);
-
-else
-   ver = hdf5read(hdf_fileName,DS.Name, DS.Attributes.Shortname);
+%mat_ver_array=datevec(version('-date'));
```
nickbattam-tessella, line 88: is this code still required?

herbert_core/utilities/hdf_nexus/find_root_nexus_dir.m
```diff
@@ -61,26 +72,28 @@
     return;
 end
 if(n_nx_entries>1)
-    error('ISIS_UTILITES:invalid_argument',' found multiple nxspe folders in file %s but this function does not currently support multiple nxspe folders',hdf_fileName)
+    if ~test_mode
```
nickbattam-tessella, line 75: why is this support implemented in test mode but not production mode?

herbert_core/utilities/hdf_nexus/find_root_nexus_dir.m
abuts: there is thought on checking it agains old matlab, but probably not worth it. Remved.

herbert_core/utilities/hdf_nexus/find_root_nexus_dir.m
abuts: It was discussion, when nxspe file format was introduced, about the possibility to store more then one nxspe data block in a file. This was not rejected but was not implemented. May be we may want to enable this possibility for our files (nxsqw?) In this case, this option should be enabled and tested more carefully but meanwhile -- let it be there but disabled.

_test/test_data_loaders/test_loader_utilites.m
```diff
@@ -44,25 +44,25 @@ function test_correct_Folder_found(this)
        function test_correct_DSinCorrectFolder_found(this)
             [DS_info,ds_path] = find_dataset_info(f_name(this,'MAP11014.nxspe'),'data','data'); 
             
-            assertEqual(DS_info.Dims,[30,28160]);
+            assertEqual(DS_info.Dataspace.Size,[30,28160]);
```
nickbattam-tessella, line 47: where did this API change come from?

abuts: hdf5info , which Matlab claims will be removed soon and h5info which is current (and is here for a while) return different types of structures. I took this ocasion to make the change.

To be fair, h5info returns more sensible outputs, so It looks like for the better

PR #160, opened 2020-05-18
Bug in custom serialization and finish_task_.m

abuts:
That's additional bug found in custom serialization code


abuts:
The bugs were discovered in Horace. More Herbert tests are needed.

PR #165, opened 2020-05-27
Package MSMPI runtime with Herbert

### LINKED ISSUE #91: Package MSMPI with Herbert
Ensure the Herbert distribution package includes MSMPI and that this MPI is built against and used by Herbert. I've checked the licence ([license_sdk.zip](https://github.com/pace-neutrons/Herbert/files/4293186/license_sdk.zip)), and we're OK to re-distribute the binaries.

We'll need to experiment a little with versions and see what does and doesn't work. Hopefully using v10 should work fine.

**Outputs:**
- MSMPI should be downloaded/already present within Herbert at build time, so the library can be built against.
- The `mpiexec` executable should be shipped and all Windows `mpiexec` calls should use it

hsautessella:
This PR puts our MSMPI build dependencies in the repo, this includes the libraries, includes and runtime. When we package Herbert, we will ship the runtime within the `herbert_core/DLL` directory.

Fixes #91 
Fixes https://github.com/pace-neutrons/Horace/issues/142


herbert_core/DLL/.gitignore
```diff
@@ -0,0 +1 @@
+mpiexec*
```
nickbattam-tessella, line 1: what is this ignoring?

CMakeLists.txt
```diff
@@ -46,20 +43,12 @@ add_subdirectory("cmake")
...
     PATTERN "*.asv" EXCLUDE
-    PATTERN "${EXTERNAL_IGNORE_PATTERN}" EXCLUDE
-)
-install(
-    FILES "LICENSE" "README.md"
-    DESTINATION "."
-)
+    PATTERN "*.gitignore" EXCLUDE
+    PATTERN "${EXTERNAL_IGNORE_PATTERN}" EXCLUDE)
+install(FILES "LICENSE" "README.md" DESTINATION ".")
```
nickbattam-tessella, line 54: Do the README or LICENSE files need updating to note that we are including the MSMPI, MPICH and HDF libraries (with license files in the appropriate `external` directory?

herbert_core/DLL/.gitignore
hsautessella: MPI runtimes are copied here on build, so they can be packaged. This ignores `mpiexec.exe` on Windows and `mpiexec` on Linux

CMakeLists.txt
hsautessella: What was the decision on this? We discussed offline that we should have a note, did we decide it should be in our licence or in the README? And what should it say exactly?

nickbattam-tessella:
Added do-not-merge tag; the test added in b83bc0a is failing

PR #176, opened 2020-06-19
bug in tmpdir on idaaas

abuts:
The issue user experienced was that if tmp folder on idaaas does not exist, Horace initialization was failing. I think, the solution is straightforward -- create such folder.

### DIFF
--- a/herbert_core/admin/tmp_dir.m
+++ b/herbert_core/admin/tmp_dir.m
@@ -16,7 +16,7 @@ if is_idaaas()
         location = getenv('HOME');
     end
     the_dir = fullfile(location,'tmp');
-    if ~exist(the_dir,'dir') == 7
+    if ~(exist(the_dir,'dir') == 7)
         [ok,the_dir,mess] = try_to_create_folder(location,'tmp');
         if ~ok
             warning('TMP_DIR:runtime_error',...
--- a/herbert_core/admin/tmp_dir.m
+++ b/herbert_core/admin/tmp_dir.m
@@ -6,9 +6,6 @@ function the_dir= tmp_dir()
 % tmp_dir     tempdir value on any machine  and (userpath()/tmp)
 %            (usually /home/user_name/Documents/MATLAB/tmp) folder if the machine is
 %            identified as iDaaaaS machine.
-%
-% $Revision:: 840 ($Date:: 2020-02-10 16:05:56 +0000 (Mon, 10 Feb 2020) $)
-%
 
 if is_idaaas()
     location = userpath();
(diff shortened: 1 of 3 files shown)


herbert_core/utilities/misc/try_to_create_folder.m
```diff
@@ -0,0 +1,38 @@
...
+
+ic = 0;
+if ~exist(folder_path,'dir')
+    [success, mess] = mkdir(folder_path);
+    while ~(exist(folder_path,'dir') || ic<3)
+        [success, mess] = mkdir(folder_path);
+        ic = ic + 1;
+        pause(0.1);
+    end
+else
```
nickbattam-tessella, line 26: what is this else branch doing? if the folder doesn't exist it tries to create it; if it does exist it tried to create randomly named path and immediately delete it

herbert_core/admin/tmp_dir.m
```diff
@@ -6,19 +6,25 @@
 % tmp_dir     tempdir value on any machine  and (userpath()/tmp)
 %            (usually /home/user_name/Documents/MATLAB/tmp) folder if the machine is
 %            identified as iDaaaaS machine.
-%
-% $Revision:: 840 ($Date:: 2020-02-10 16:05:56 +0000 (Mon, 10 Feb 2020) $)
-%
 
 if is_idaaas()
```
nickbattam-tessella, line 10: this `if` makes the code ISIS specific. Is there any way to use a standard approach across ALL platforms so we don't need this site-specific code branch?

herbert_core/utilities/misc/try_to_create_folder.m
```diff
@@ -0,0 +1,38 @@
...
+    test_path = fullfile(folder_path,['folder_twa_',char(randi(25,1,10) + 64)]);
+    %clob = onCleanup(@()rmdir(test_path));
```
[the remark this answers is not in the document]
abuts: probably yes, with hard feeling and  re-estimating the previous experience. There is the need to delete this temporary folder and comment is the safest way of doing that. it does not fit other useful code below, so commented....

herbert_core/utilities/misc/try_to_create_folder.m
abuts: yes -- its the part of the code, which creates folders for configuration (I've reused existing piece of code for tmp folder.)
The issue is to ensure that the folder have correct write access. 
Apparently, creating something is simplest and most reliable way of checking write access on any system.

abuts: delay and repetitive attempts to create come from Windows, where in some configurations something (looks like antivirus (or virus?) temporary locks some folders.

herbert_core/admin/tmp_dir.m
abuts: can not think of one. The iDaaaS issue is that two different machines (large and small) with different Horace configurations work from one users folder. There are also other strange issues we need to deal with, like this tmp directory, which can not be normal tmp directory as on normal OS. this is why is_idaaas was introduced and needs to be maintained. We also have idaaas specific configurations.

abuts:
@hsautessella  -- I've already merged the branch into release_3.4 as this is the bugfix to test with user. Alternative would be to fix release and cherrypick commits to master (or creating branch from cherrypicking) -- IMHO -- headache....

This is safe change, as it is very unlikely that bug in this change affect other users. For complex changes, where other users may be affected or situation is unclear, we need to create a copy of the release code tree, and switch user to this copy until code is fixed, tested and can be safely merged to the main tree. This task needs interaction with iDaaaS team and is on my TODO list.

PR #177, opened 2020-06-19
Release PR

abuts:
Formal PR to check the validity of the release as bugfix has been deployed
I believe, should not be merged under standard workflow, though no harm in doing that in this particular situation.


abuts:
Completely agree -- we should not merge such branches into master. Vise versa  shuld usually be safe though not requested in this case.

PR #180, opened 2020-06-25
Add help to build scripts

### LINKED ISSUE #179: Add help to build scripts
Herbert equivalent to https://github.com/pace-neutrons/Horace/issues/257

hsautessella:
This PR adds help to Herbert's `build.ps1` and `build.sh` build scripts. The help messages should be printed given the `-h` or `-help` flags in PowerShell and  the -h` or `--help` flags in bash. Help should also be printed if no arguments are passed to the help scripts.

Refs: https://github.com/pace-neutrons/Horace/issues/257

Fixes #179

### DIFF
--- a/tools/build_config/build.ps1
+++ b/tools/build_config/build.ps1
@@ -25,19 +25,19 @@
   https://github.com/pace-neutrons/Herbert
 #>
 param (
-  # Run the Herbert build commands
+  # Run the Herbert build commands.
   [switch][Alias("b")]$build,
-  # Run all Herbert tests
+  # Run all Herbert tests.
   [switch][Alias("t")]$test,
-  # Pacakge Herbert into a .zip file
+  # Pacakge Herbert into a .zip file.
   [switch][Alias("p")]$package,
-  # Print the versions of libraries being used e.g. Matlab
+  # Print the versions of libraries being used e.g. Matlab.
   [switch][Alias("v")]$print_versions,
-  # Call Get-Help on this script and exit
+  # Call Get-Help on this script and exit.
   [switch][Alias("h")]$help,
 
-  # The version of Visual Studio to build with (one of 2015, 2017, 2019).
-  # Other Windows compilers are not supported by this script. {2015, 2017, 2019}
+  # The version of Visual Studio to build with. Other Windows compilers are
+  # not supported by this script. {2015, 2017, 2019}
   [int][ValidateSet(2015, 2017, 2019)]
   [Alias("VS")]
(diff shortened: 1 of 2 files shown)


tools/build_config/build.ps1
```diff
@@ -1,29 +1,70 @@
...
   [Alias("C")]
   $build_config = 'Release',
 
+  # The directory to write build files into. If the directory does not exist it
+  # will be created.
   [string]
   [Alias("O")]
   $build_dir = "",
 
+  # Flags to pass to the CMake configure step.
```
nickbattam-tessella, line 62: inconsistent use of `.` at the end of these help strings

tools/build_config/build.ps1
```diff
@@ -1,29 +1,70 @@
...
   [switch][Alias("t")]$test,
+  # Pacakge Herbert into a .zip file
   [switch][Alias("p")]$package,
+  # Print the versions of libraries being used e.g. Matlab
   [switch][Alias("v")]$print_versions,
+  # Call Get-Help on this script and exit
+  [switch][Alias("h")]$help,
 
+  # The version of Visual Studio to build with (one of 2015, 2017, 2019).
+  # Other Windows compilers are not supported by this script. {2015, 2017, 2019}
```
nickbattam-tessella, line 40: is there a standard way to show the default in the parameter help?

hsautessella: You need to do `Get-help ./build.ps1 -Full` then the parameter output is a bit more verbose:
```powershell
    -build [<SwitchParameter>]
        Run the Herbert build commands.

        Required?                    false
        Position?                    named
        Default value                False
        Accept pipeline input?       false
        Accept wildcard characters?  false

    -test [<SwitchParameter>]
        Run all Herbert tests.

        Required?                    false
        Position?                    named
        Default value                False
        Accept pipeline input?       false
        Accept wildcard characters?  false
```

I'm not sure what you think is best here. Should I also include the default in the help string? It's more to manually maintain, but is more obvious from a user's perspective

nickbattam-tessella: I think it would be worth it, that's a lot of extra clarity for a relatively small maintenance overhead and the default is defined in the same block of code a couple of lines below

Simple appending `... [ default-value ]` to the text would work e.g.
```
 # Run all Herbert tests [false]
[switch][Alias("t")]$test,
```
For enumerated options `..... {2015, 2017, 2019} [2017]` still parses reasonably clearly

PR #185, opened 2020-07-07
minor changes improving compartibility with Maltab 2015

abuts:
This is small change to improve Herbert compatibility with previous Matlab versions. No task to achieve full compatibility is tried or achieved, but the changes are still useful to for the future.


nickbattam-tessella:
This change leaves the code less readable, and is only achieving partial compatibility.

Is there any value in merging it? We cannot guarantee there will not be further compatibility breaking changes introduced into Herbert and Horace as we're not targeting it support.

abuts:
It does not makes it less reliable.  Old code works and was working for ages. 
I do not have time to get full reliability but it 1) can be asked by special peoples. 2) The way to check one idea why long parallel job currently hangs up.

PR #191, opened 2020-07-11
job folder migration

abuts:
This fix should eliminate failures between jobs


herbert_core/classes/MPIFramework/parallel_worker.m
```diff
@@ -295,7 +298,9 @@
     fbMPI = je.control_node_exch;
     % is framework instance different from JE instance now?
     intercomm= je.mess_framework;
-    
+    % migrate job folder for message exchange without deleting the old
+    % one
+    je.migrate_job_folder(false);
```
nickbattam-tessella, line 303: why is the folder migrated, but not cleaned up in this use-case?

herbert_core/classes/MPIFramework/parallel_worker.m
```diff
@@ -273,6 +273,9 @@
                 fbMPI = je.control_node_exch;
                 % is framework instance different from JE instance now?
                 intercomm= je.mess_framework;
+                % migrate job folder for message exchange without deleting the old
+                % one
+                je.migrate_job_folder(false);
```
nickbattam-tessella, line 278: why is the folder migrated, but not cleaned up in this use-case?

herbert_core/classes/MPIFramework/@ClusterWrapper/ClusterWrapper.m
```diff
@@ -258,6 +266,10 @@
             % and complete parallel job
             if ~isempty(obj.mess_exchange_)
                 obj.mess_exchange_.finalize_all();
+                new_mess_exchange_folder = obj.mess_exchange_.next_message_folder_name;
+                if exist(new_mess_exchange_folder,'dir')==7
+                    rmdir(new_mess_exchange_folder,'s');
```
nickbattam-tessella, line 271: can this throw an exception and bring down the framework wrapper?

herbert_core/classes/MPIFramework/parallel_worker.m
abuts: the workers and the server itself check the presence of the exchange folder and if it is deleted, fail. To ensure this would not happen, for workers during migration, I have done changes to JobExecutor.finish_task. But server still checks the folder existence and if not find one, fails itself. This is why I've dedicated the honor of deleting job folder to server only

abuts: the migration not only for failures. The normal job may still leave some garbage that will confuse the next job and may have issues with startup files cleared up by the previous job (hopefully resolved earlier)

herbert_core/classes/MPIFramework/parallel_worker.m
abuts: ibid (see above)

herbert_core/classes/MPIFramework/@ClusterWrapper/ClusterWrapper.m
abuts: this is finalize procedure -- it happens when framework is shut down anyway. Especially when user types CTRL-C

herbert_core/classes/MPIFramework/parallel_worker.m
abuts: In addition to that, the server may not pick up a client outputs, so clients should not delete exchange folder

Issue #192, opened 2020-07-15
Rundata serialization


abuts:
This ticket is Herbert part of Horace ticket [#287](https://github.com/pace-neutrons/Horace/issues/287) -- symmetrization does not work in parallel mode. The reason for this, apparently in the fact that the symmetrization function is not serialized/deserizlized correctly.

PR #195, opened 2020-07-15
optimal config

abuts:
Minor changes to modify iDaaaS and isiscompute default configurations


herbert_core/classes/MPIFramework/@MessagesFilebased/MessagesFilebased.m
```diff
@@ -26,7 +26,7 @@
     end
     properties(Constant=true,Access=protected,Hidden=true)
         % The number, used in folder name and Job ID to define folder migration.
-        folder_migration_shift =  0011100001;
+        folder_migration_shift =  0000000001;
```
nickbattam-tessella, line 29: What is the relevance of this bitmask change?

abuts: that's have actually picked up from another ticket -- nothing to do with config 
this one is e0f1b72 only

abuts: To be fair, I've started this one from #192 by mistake, though all should go in the same place.

PR #199, opened 2020-07-17
try catch in worker

abuts:
Small change which seems does not make much sense but can be useful.


admin/worker_4tests.m.template
```diff
@@ -13,14 +13,39 @@ function [ok,err_mess,je]=worker_4tests(worker_controls_string)
...
-% Used for filebased framework debugging on windows, when one can get access to the terminal of running worker.
-[ok,err_mess,je] = parallel_worker(worker_controls_string,DO_LOGGING,DO_DEBUGGING);
+if ~ok && isa(err_mess,'MException')
+    if isa(je,'JobExecutor')
+        log_file = fullfile(tempdir,sprintf('WORKER_V2_Node#%d_failure.log',je.labIndex));        
+    elseif isempty(je)
+        log_file = fullfile(tempdir,'WORKER_V2_INIT_failure.log');
+    end
+    
+    fh = fopen(log_file,'w');
```
nickbattam-tessella, line 45: this inner file open/write/close block needs to be in a nested try...catch, otherwise if/when it fails it will leave an uncaught exception

admin/worker_v2.m.template
```diff
@@ -13,8 +13,36 @@ function [ok,err_mess,je]=worker_v2(worker_controls_string)
...
+        log_file = fullfile(tempdir,'WORKER_V2_JE_FAIL_PROCESSING_failure.log');
+    end
+    err_mess = ME1;
+end
+if ~ok && isa(err_mess,'MException')
+    if je == 1  % unhandled exception at initialization procedure. No Job_ID labNum available
+        log_file = fullfile(tempdir,'WORKER_V2_INIT_failure.log');
+    end
+    
+    fh = fopen(log_file,'w');
```
nickbattam-tessella, line 42: this inner file open/write/close block needs to be in a nested try...catch, otherwise if/when it fails it will leave an uncaught exception

herbert_core/classes/MPIFramework/parallel_worker.m
```diff
@@ -175,11 +180,18 @@
         % something wrong with the code. We can not process interrupt
         % properly, but filebased framework should still be
         % available.
-        if DO_LOGGING; log_input_message_exception_caught();  end
-        err_mess = sprintf('job N%s failed. Error during job initialization %s:',...
-            control_struct.job_id,ME.message);
-        fbMPI.send_message(0,FailedMessage(err_mess,ME));
-        break;
+
+        if ~strcmp(ME.identifier,'MESSAGE_FRAMEWORK:canceled')
```
nickbattam-tessella, line 184: possible typo: `MESSAGE_FRAMEWORK:canceled` or `MESSAGE_FRAMEWORK:cancelled` ?

admin/worker_4tests.m.template
```diff
@@ -13,14 +13,39 @@ function [ok,err_mess,je]=worker_4tests(worker_controls_string)
...
 %              folder name where the job initialization data are located on
 %              a remote system.
-
+je = [];
 if isempty(which('herbert_init.m'))
     herbert_on();
-    % Set up non-default herbert or Horace path to test them
-    % herbert_on('/home/isis_direct_soft/Herbert_bugfix/herbert_core');
-    % horace_on('/home/isis_direct_soft/Horace_bugfix/horace_core');
+    %horace_on('c:\Users\abuts\Documents\developing_soft\Horace_3_4\horace_core');
```
hsautessella, line 19: Comment left here.

Also, this `herbert_on` call should be within the `try-catch` below. If `herbert_on` fails we could end up leaving processes lying around

admin/worker_v2.m.template
```diff
@@ -13,8 +13,36 @@ function [ok,err_mess,je]=worker_v2(worker_controls_string)
...
+DO_LOGGING = false;
+try
+    [ok, err_mess,je] = parallel_worker(worker_controls_string,DO_LOGGING);
+catch ME1 % intercepted exception in processing failure or some odd bug indeed
+    ok = false;
+    if exist('je','var') && isa(je,'JobExecutor') %exception thrown but JE is defined
+        if isnumeric(je.labIndex) && ~isempty(je.labIndex)
+            log_file = fullfile(tempdir,sprintf('WORKER_V2_Node#%d_failure.log',je.labIndex));
+        else
+            log_file = fullfile(tempdir,sprintf('WORKER_V2_Node_JE_common_failure.log'));
```
hsautessella, line 30: Why not use a process ID here (`feature('getpid')`)? Then, if we get multiple failures, the processes are not all trying to write to the same file

herbert_core/classes/MPIFramework/parallel_worker.m
abuts: no -- its indeed American spelling -- it in so many places, that better leave it alone.

admin/worker_v2.m.template
abuts: good point -- not sure if it works but worth trying

herbert_core/classes/MPIFramework/parallel_worker.m
nickbattam-tessella: just wanted to check

admin/worker_4tests.m.template
abuts: done

PR #201, opened 2020-07-17
Build mex libraries directly into DLL directory

### LINKED ISSUE #200: Build mex files directly into DLL directory
At the moment, the generated mex files are built in `build/bin` and then copied into the `DLL` directory. This can raise one or two issues such as, if you delete the mex libraries from the DLL directory and re-build, the libraries are not re-copied. Building directly into `DLL` also ensures any debugging files (e.g. pdb files) are put in the correct place.

hsautessella:
They were previously generated in build/bin and copied across. Doing it directly reduces duplication, makes sure debugging files (e.g. .pdb files) are put in the correct place and makes sure CMake knows when it needs to re-build a target.

Fixes #200


hsautessella:
Should not be merged until https://github.com/pace-neutrons/Horace/pull/295 has been

cmake/shared/PACE_AddMex.cmake
```diff
@@ -85,15 +78,11 @@ function(pace_add_mex)
...
-    if(${prefix}_COPY_TO)
-        set(_target_file "$<TARGET_FILE:${${prefix}_NAME}>")
-        set(_dest_file "${${prefix}_COPY_TO}/$<TARGET_FILE_NAME:${${prefix}_NAME}>")
-        add_custom_command(TARGET "${${prefix}_NAME}"
-            POST_BUILD
-            COMMAND ${CMAKE_COMMAND} -E copy "${_target_file}" "${_dest_file}"
-            COMMENT "Copying ${${prefix}_NAME}.${Matlab_MEX_EXTENSION} into ${${prefix}_COPY_TO}"
+    set_target_properties("${${prefix}_NAME}"
+        PROPERTIES
+            # Appending the $<0:> generator expression prevents MSVC outputing
```
nickbattam-tessella, line 83: Rather than saying what it prevents, this comment should say what it causes to happen, i.e. build artifacts are placed in the XX_OUTPUT_DIRECTORY folder rather than the normal Release or Debug folder

PR #208, opened 2020-07-21
iDaaaS bugs

abuts:
as per ticket

### DIFF
--- a/herbert_core/utilities/global_path/resolve_path.m
+++ b/herbert_core/utilities/global_path/resolve_path.m
@@ -1,6 +1,10 @@
-function fullpath = resolve_path(filename)
+function [fullpath,ok] = resolve_path(filename)
 % Function to resolve full path to a file or a folder
 %
+% Returns:
+% Full path to the file
+% ok -- if file exist and can be resolved, ok==true
+%       if false, file not exist
 %
 if ~ischar(filename)
     error('RESOLVE_PATH:invalid_argument',...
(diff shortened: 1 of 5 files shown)


herbert_core/utilities/global_path/resolve_path.m
```diff
@@ -0,0 +1,23 @@
...
+%
+if ~ischar(filename)
+    error('RESOLVE_PATH:invalid_argument',...
+        'input should be a string, identifying file or path to a file. Got: %s',...
+        fevalc('disp(filename)'))
+end
+
+file=java.io.File(filename);
+fullpath = char(file.getCanonicalPath());
+if isunix && fullpath(end) == '~'
```
nickbattam-tessella, line 13: this doesn't correctly handle `~` in Linux:
```
>>> cd /home/user/bob/workingDir
>>> g = java.io.File("~/test.file");
>>> g.getCanonicalPath()  # should be /home/user/bob/test.file
/home/user/bob/workingDir/~/test.file
```
The `~` should be replaced by user homedir

nickbattam-tessella: Since `~` will always start the path you can check and switch before creating the file:
```
if starts_with_tilde
  file = java.io.File(userhome, filename(2:end))
else
  file = java.io.File(filename)
end
```

abuts: Also does not work. 
In addition to that,  getuserpath on iDaaaS resolves to simulink and java(~) to real mounting path. Theoretically the same, but pain for unit tests

PR #212, opened 2020-08-05
standard api for cpp communicator

### LINKED ISSUE #155: CPPcommunicator does not receives persitent messages
Persistent messages are the messages, describing MPI framework interrupts. In Parallel computing toolbox and filebased messages framework they are received even if the client asks for any other messages. CPPcommunicator framework does not, seems have such feature and CPPMpi framewok behaves differently. This issue should be addressed.

abuts:
This changes CPP communicator API the same as two other frameworks. 
should also fix #155


_LowLevelCode/cpp/cpp_communicator/MPI_wrapper.h
```diff
@@ -70,13 +78,21 @@ class MPI_wrapper {
...
+    SendMessHolder *get_interrupt_queue(int dest_address = 0) {
+        return &this->InterruptHolder[dest_address];
     }
     // check if any message present in test mode
     bool any_message_present() {
-        if (SyncMessHolder.theRequest==0) 
-            return true;
+        for (auto it = InterruptHolder.begin(); it != InterruptHolder.end(); it++) {
+            if (it->theRequest == 0)  return true;
+        }
```
hsautessella, line 92: The standard way of doing this would be with a range-based for loop:

```suggestion
        for (auto &msg_holder : InterruptHolder) {
            if (msg_holder.theRequest == 0)  
                return true;
        }
```

_LowLevelCode/external/CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ if(UNIX)
 else()
     set(Herbert_MSMPI_DIR "${CMAKE_CURRENT_LIST_DIR}/win64/MSMPI-8.0.12")
     set(Herbert_MPIEXEC_EXE "${Herbert_MSMPI_DIR}/Bin/mpiexec.exe")
+    set(Herbert_MPIEXEC_EXE "${Herbert_MSMPI_DIR}/Bin/smpd.exe")
+    set(Herbert_MPIEXEC_EXE "${Herbert_MSMPI_DIR}/Bin/msmpilaunchsvc.exe")
```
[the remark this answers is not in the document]
abuts: yes, this does not work. Unfortunately, much more is not working there now

_test/test_mpi/test_job_dispatcher_mpiexec.m
```diff
@@ -19,27 +19,27 @@ function test_ClusterMPI_get_mpiexec_executable_exists(~)
...
         function test_job_fail_restart(obj, varargin)
             % DISABLED on Windows -- random hang-up of one or another
-            if ispc
-                warning('test_job_fail_restart disabled on Windows')
-                return;
-            end
+%             if ispc
+%                 warning('test_job_fail_restart disabled on Windows')
+%                 return;
+%             end
```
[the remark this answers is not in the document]
abuts: yes, this should

herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
```diff
@@ -196,11 +195,12 @@ function check_availability(obj)
     methods(Static)
         function mpi_exec = get_mpiexec()
             rootpath = fileparts(which('herbert_init'));
-            external_dll_dir = fullfile(rootpath, 'DLL', 'external');
+            %external_dll_dir = fullfile(rootpath, 'DLL', 'external');
+            external_dll_dir = fullfile(rootpath, 'DLL');
```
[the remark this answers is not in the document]
abuts: that's part of the issues why its failing and I am investigating it now

_LowLevelCode/cpp/cpp_communicator/MPI_wrapper.h
abuts: may be. Hate references from FORTRAN times. Does this mean that every variable in iteration is copied into `msg_holder` or not? Its not important in this case, as this code is for tests only, but hate using the code, not fully understanding what is happening

hsautessella: No, in this case nothing would be copied. `msg_holder` will simply point to the items in `InterruptHolder`.

The `&` means that nothing is copied.

```cpp
for (auto msg_holder : InterruptHolder) { ... }  // msg_holder copies items in InterruptHolder
for (auto &msg_holder : InterruptHolder) { ... }  // msg_holder points to items in InterruptHolder
```

hsautessella: In fact, since we're not editing `msg_holder` the loop should really be `for (const auto &msg_holder) { ... }`. I've changed the suggestion above to reflect this.

nickbattam-tessella: Can we please have brackets around the `return true;` in the `if` too

_LowLevelCode/cpp/cpp_communicator/cpp_communicator.cpp
```diff
@@ -190,10 +199,10 @@ void mexFunction(int nlhs, mxArray* plhs[], int nrhs, const mxArray* prhs[])
 /* If appropriate number of output arguments are availiable, set up the mex routine output arguments to mpi_numLab and mpi_labNum values
    extracted from initialized MPI framework.
 */
-void set_numlab_and_nlabs(class_handle<MPI_wrapper> const* const pCommunicatorHolder, int nlhs, mxArray* plhs[], int nrhs, const mxArray* prhs[]) {
+void set_numlab_and_nlabs(class_handle<MPI_wrapper> * const pCommunicatorHolder, int nlhs, mxArray* plhs[], int nrhs, const mxArray* prhs[]) {
 
     if (nlhs >= (int)labIndex_Out::numLab + 1) {
-        plhs[(int)labIndex_Out::numLab] = mxCreateNumericMatrix(1, 1, mxUINT64_CLASS, mxREAL);
+        plhs[(int)labIndex_Out::numLab] = mxCreateNumericMatrix(1, 1, mxINT32_CLASS, mxREAL);
```
nickbattam-tessella, line 205: why switch from `UINT64` to `INT32` here?

_LowLevelCode/cpp/cpp_communicator/input_parser.h
```diff
@@ -144,17 +152,17 @@ mxArray * class_handle<T>::export_hanlder_toMatlab()
     *pData = reinterpret_cast<uint64_t>(this);
     return out;
 }
-
+//
```
nickbattam-tessella, line 155: empty comment

herbert_core/classes/MPIFramework/@iMessagesFramework/iMessagesFramework.m
```diff
@@ -722,16 +723,24 @@ function set_job_id_(obj,new_job_id)
             if ~isnumeric(from_task_id)
                 error('MESSAGES_FRAMEWORK:invalid_argument',...
                     'Task_id to receive message should be a number');
-            elseif numel(from_task_id)>1
-                error('MESSAGES_FRAMEWORK:invalid_argument',...
-                    'Receiving only one message from one lab may be requested. Asked for: %d',...
-                    numel(from_task_id));
+            else
```
nickbattam-tessella, line 726: why is this an `else` with two nested `if`s rather than 
```
...
elseif numel(from_task_id)>1
...
elseif from_task_id>obj.numLabs
...
end
```

herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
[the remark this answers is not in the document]
abuts: changed in search for error. Fixed.

_LowLevelCode/cpp/cpp_communicator/MPI_wrapper.h
```diff
@@ -19,38 +19,43 @@ class SendMessHolder {
     int destination;
     // vector of the message contents, used as the buffer of the message contents until the message is received
     std::vector<uint8_t> mess_body;
-      //SendMessHolder(SendMessHolder&& other) noexcept;
+    //SendMessHolder(SendMessHolder&& other) noexcept;
```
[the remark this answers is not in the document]
abuts: let it still be there -- I do not understand why the move constructor does not work. Will delete it at completion.

_LowLevelCode/cpp/cpp_communicator/cpp_communicator.cpp
abuts: because inputs become 32 -- I think the number of parallel workers  of  2^31 order is well sufficient for any imaginable purpose

_LowLevelCode/cpp/cpp_communicator/MPI_wrapper.h
abuts: not sure its entirely true because 
`
   for (auto &&it : InterruptHolder) {
            if (it.theRequest == 0)  return true;
        }
`
works absolutely fine too. And its not a pointer.... Will try const 
@nickbattam-tessella  -- I've changed it to one-liner -- looks better without the brackets.

hsautessella: I wasn't completely accurate with what I said before `for (auto &msg_holder : InterruptHolder)` doesn't make `msg_holder` a `SendMessHolder`, it's an iterator (a `std::vector<T>::iterator`, which is what `InterruptHolder.begin()` returns). 

The difference between `for (auto &x : vec ) {}` and `for (auto &&x : vec) {}` is explained here https://stackoverflow.com/questions/29320575/whats-the-difference-between-and-in-a-range-based-for-loop

_LowLevelCode/cpp/cpp_communicator/input_parser.h
abuts: fiext

_LowLevelCode/cpp/cpp_communicator/MPI_wrapper.h
abuts: though, according to the article, most performant way of doing requests would be && as its direct address of the data. 
const & may default to iterator.

herbert_core/classes/MPIFramework/@MessagesCppMPI/MessagesCppMPI.m
```diff
@@ -221,6 +221,16 @@ function delete(obj)
             %       failure.
             % >> on success, message contains an object of class aMessage,
             %        with the received message contents.
+            %NOTE:
+            % When synchronous receive is selected, receive waits for
+            % the specific message with specific data tag to be send.
+            % If interrupt message appears after the framework starts
+            % waiting for a synchronous message, the framework hangs up.
+            % Receive_all should be used to avoid such hang ups.
```
nickbattam-tessella, line 229: Is there any use case for not using `receive_all` if as you describe here the framework will hang up if any async message is received while waiting for a blocking one?

If there isn't should these non-`receive-all` methods simply be removed or made private? Having an API that we *know* will cause the framework to fail feels like a high risk strategy.

_LowLevelCode/cpp/cpp_communicator/MPI_wrapper.h
hsautessella: `auto &&` does not give `const` correctness. If you only want read-access to the items in the container, you should use `const`

herbert_core/classes/MPIFramework/@MessagesCppMPI/MessagesCppMPI.m
abuts: as discussed, only interrupt will hang up the framework as no data message will be send after it. An asynchronous message will be successfully received during next appropriate receive operation, so this behavior constrains only error processing. But this is weak constrain, slightly restricting transmission of the debug information, so I think this behavior should be kept as important optimization tool.

Issue #178, opened 2020-06-23
Modify CPP mpi framework and enable DataMessages


abuts:
Matlab MPI and Herbert MPI frameworks are currently have common interface and behave similarly.
The same changes and substantial part of Matlab code should be adapted for the tests and CPP MPI framework working the same way as two other frameworks. 

Related to #126 in the CPP MPI part.
Depends on #155

PR #214, opened 2020-08-07
data messages

abuts:
This PR implements data messages and finishes local mpiexec based parallel framework


_test/test_mpi_wrappers/test_exchange_CppMPI.m
```diff
@@ -66,108 +66,6 @@ function test_MessagesCppMPI_constructor(obj)
         %-----------------------------------------------------------------
         % CURRENTLY DISABLED
         %
```
hsautessella, line 68: Do we need this comment anymore?

abuts: that's placeholder left in process to see if anything still needs to be disabled. Apparently not. Removing

abuts:
fixed #178

PR #220, opened 2020-09-07
Update multifit tests to use new syntax, but independently retaining …

tgperring:
…legacy multifit and fit tests. A couple of big fixes in mfclass made as well.


abuts (approved):
does the job and looks completely fine to me

nickbattam-tessella (approved):

PR #221, opened 2020-09-08
Move Jenkins pipeline param declaration into Jenkinsfile

### LINKED ISSUE #219: Create Release CI pipeline
Herbert version of https://github.com/pace-neutrons/Horace/issues/331

hsautessella:
This PR moves the declaration of pipeline parameters into the Jenkinsfile. This means we shouldn't need to do as much configuration within the Jenkins GUI.

When parameters are set in the Jenkinsfile, the pipeline's configuration will update according to the parameters specified. If running from a clean pipeline (with no configuration set in the GUI) the first build will most likely fail because the parameters were not previously set. This should not be a problem for us since the parameters I've added in the Jenkinsfile are already set in the GUI, therefore builds should be unaffected.

The new Jenkinsfile deduces what the default parameter values should be using the pipeline's name. If the pipeline's name starts with `PR-` then the Jenkinsfile will set the parameter defaults that are required by the pull request pipelines. This equally applies to the Matlab version and OS.

The Jenkins SMG will need updating when the Horace equivalent of this PR is opened.

These changes and the creation of the `Release-*` pipelines in the Jenkins GUI fixes #219 

Fixes https://github.com/pace-neutrons/Horace/issues/346

### DIFF
--- a/tools/build_config/Jenkinsfile
+++ b/tools/build_config/Jenkinsfile
@@ -1,13 +1,113 @@
 #!groovy
-/**
- * The following parameters must be provided by the Jenkins build job and must match available application
- * on the build target machine
- *
- *  - CMAKE_VERSION: The version of CMake to run the build with
- *  - GCC_VERSION: The version of GCC to build with
- *  - MATLAB_VERSION: The release number of the Matlab to load
- *  - CPPCHECK_VERSION: The version of CppCheck tooling to load to provide the code-style checks
- */
+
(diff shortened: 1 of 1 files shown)


hsautessella:
Pipelines using the updated Jenkinsfile have been tested on pipelines such as Branch-Windows-10-TESTER-2018b. These pipelines have since been deleted.

nickbattam-tessella (approved):

PR #223, opened 2020-09-09
added common test function to the path if Horace is present

### LINKED ISSUE #222: Herbert does not add Horace common (test) functions when tests are enabled
minor bug. I wander how we are running Horace unit tests without it.

abuts:
fixes  #222. Trivial fix


nickbattam-tessella:
Why does Herbert need access to Horace common test files? Herbert should not be accessing anything in Herbert

hsautessella (commented):
I agree with Nick that Herbert shouldn't know anything about Horace. Herbert is a dependency of Horace, not the other way around.

Would adding an `init_tests` option to `hor_config` be a better idea?

There's quite a lot of duplication within `_test/common_functions` in Herbert and Horace, which is probably why things work at the moment. If we're now adding both folders to the path for tests, doesn't that raise ambiguity over which functions we're actually calling: it could be the version in `Horace/_test/common_functions` or the version in `Herbert/_test/common_functions`.

abuts:
The problem of Horace dependency on Herbert is very similar to the issue of a function called on a parent class pointer redefined in a child class. Common solution is virtual or abstract functions and life will be extremely difficult without them. 

The similar technique on our packages sounds a bit dodgy, but may be implemented (a function put on a path later overrides the function with the same signature defined before)

Not sure it worth efforts though I am more and more thinking about it....
init_tests in Horace will not work as separate Herbert build will fail.
moving common functions would work, but again, it will be strange  to have in Herbert functions, depending on Horace. 

As addition of horace top admin folder in Herbert is happening anyway, its reasonable to add common functions in this block too. 

I think, it can be removed if horace_git/admin is moved from herbert_init/herbert_config e.g. in a way, described above but it is clearly independent ticket

PR #236, opened 2020-09-23
fix noisify bugs #234, extend interface for paging

### LINKED ISSUE #234: Noisify returns std deviations, not variances, and incorrectly checks for empty input error.
General consensus is that the error part of [signal, error] in PixelData is a variance.

Currently noisify (herbert_core\utilities\math) returns standard deviations, by constructing variances for the noise added to the signal and then taking their square root (LL.52, 57). Additionally at L.52, the input variance is treated as a standard deviation and is squared before combining with the noise contribution.

Additional problem with this function is that at L.50 the input error is checked for existence and for being empty; at L.51 the sizes of the input and noise errors are checked to be equal. This looks like the wrong way round - the check should be against NOT empty.

cmarooney-stfc:
This PR does the following:

- it does fix #234 to correct 
    - (i) interpretation of the input e and output eout as standard deviations rather than variances - sqrts and ^2s are removed
    - (ii) isempty test on the error output processing at the end is switched to not isempty to allow correct comparison of input and output error sizes

- it also supports a fourth argument to pass a previously calculated max signal value to replace the locally calculated one. This allows a max to be calculated in a previous pass when paging is used.
    - the existing interface is unchanged
    - if the fourth argument is present, the third argument can be defaulted using a value <0.

### DIFF
--- a/herbert_core/utilities/maths/noisify.m
+++ b/herbert_core/utilities/maths/noisify.m
@@ -28,39 +28,54 @@ function [yout,eout] = noisify (y,e,varargin)
 %
 %   If no input errors, e, just set e=[]
 
-if nargin==3 && ischar(varargin{1})
-    if is_stringmatchi(varargin{1},'poisson')
+% Use Poisson distribution and ignore other arguments
+if nargin==3 && ischar(varargin{1}) && is_stringmatchi(varargin{1},'poisson')
         yout=zeros(size(y));
         for i=1:numel(y)
             yout(i)=randpoisson(abs(y(i)));
         end
         eout=abs(y);  % the input y is the mean and variance of the Poisson distribution
+        return % RETURN here as the poisson route is independent of the other argument options
+end
+
+% If not return, use normal distribution with or without an input maximum signal:
+default_fac = 0.1;
+ymax = [];
+fac = default_fac;
+if nargin>=3
+    % Check for maxval keyword and set ymax, else calc ymax locally below
+    pos = find(strcmp(varargin,'maxval')==1);
+    if ~isempty(pos) && pos<size(varargin,2)
+        ymax = varargin{pos+1};
     else
-        error('Unrecognised random sampling distribution')
+        error('Could not find maxval value arg with maxval specified')
     end
-else % 3rd arg did not exist or was not a distribution name
-    default_fac = 0.1;
-    if nargin>=3 % 3 args for fac or 4 args for fac and maxval
+    % Check for input of fac as 3rd arg, set to default otherwise
+    if isempty(pos) || pos ~= 1
         if isnumeric(varargin{1}) && isscalar(varargin{1})
             fac = varargin{1};
-            if fac < 0.0 % default option if 4th arg is needed
-                fac = default_fac;
-            end
         else
             error('Noise as fraction of peak signal must be real scalar')
         end
     else
         fac = default_fac;
     end
-    if nargin==4 % means the maxval arg has been used
-        ymax = varargin{2};             % use max value previously extracted
-    else
-        ymax = max(abs(y(:)));          % find maximum magnitude of y for arbitrary dimensions
-    end
-    dy=(fac*ymax)*randn(size(y));   % st. dev. of randn is sigma=1
-    yout=y+dy;
-    eout=ones(size(y))*(fac*ymax)^2;
 end
+
+% Check for any other distribution in arg #3
+if nargin>=3 && ischar(varargin{1}) && ~is_stringmatchi(varargin{1},'maxval')
+    error('Unrecognised random sampling distribution')
+end
+
+% if ymax was not set by an argument, set from max of |y|
+if isempty(ymax)
+    ymax = max(abs(y(:)));
+end
+    
+% make noise dy and add to y for output; make error bar for noise    
+dy=(fac*ymax)*randn(size(y));   % st. dev. of randn is sigma=1
+yout=y+dy;
+eout=ones(size(y))*(fac*ymax)^2;
 
 % this code now outputs a variance to be consistent between the use of eout
 % in the Poisson distribution section, the input definition for e and this
--- a/herbert_core/utilities/maths/noisify.m
+++ b/herbert_core/utilities/maths/noisify.m
@@ -18,13 +18,14 @@ function [yout,eout] = noisify (y,e,varargin)
 %           Add noise with Poisson distribution, where the mean value at
 %           a point is the value y.
 %
-%   >> [yout, eout] = noisify(y,e,factor,maxval)
+%   >> [yout, eout] = noisify(y,e,[factor,]'maxval',maxval)
 %           Add noise with Gaussian distribution, calculating the standard
 %           deviation by an externally provided maximum y value.
-%           Typically this will be the overall maximum value if the
+%           The max value is preceded by a keyword string 'maxval'.
+%           Typically this value will be the overall maximum value if the
 %           data is processed page by page and the maximum must be
 %           extracted before this processing.
-%           Setting factor<0.0 will use the default value 0.1.
+%           Omitting factor will use the default value 0.1.
 %
 %   If no input errors, e, just set e=[]
 
--- a/herbert_core/utilities/maths/noisify.m
+++ b/herbert_core/utilities/maths/noisify.m
@@ -33,9 +33,8 @@ function [yout,eout] = noisify(y,e,varargin)
 % deal with optional arguments
 is_poisson = false;
 fac = 0.1;
-USE_LOCAL_MAX = -999; % set default ymax to distinctive negative value
 if nargin>=2 % avoids the case where the only paramemter is y; for >=2 e will not be optional
-    [fac, is_poisson, ymax] = parse_args(y, e, fac, is_poisson, USE_LOCAL_MAX, varargin{:});
+    [fac, is_poisson, ymax] = parse_args(y, e, fac, is_poisson, varargin{:});
 end
 
 % Use Poisson distribution and ignore other arguments
@@ -47,8 +46,8 @@ if is_poisson
     eout=abs(y);  % the input y is the mean and variance of the Poisson distribution
 else
     % if ymax was not set by an argument, set from max of |y|
-    if ymax == USE_LOCAL_MAX
-        ymax = max(abs(y(:)));
+    if ymax == get_maximum_signal_value('default')
+        ymax = get_maximum_signal_value(y);
     end
 
     % make noise dy and add to y for output; make error bar for noise    
@@ -68,13 +67,13 @@ if exist('e','var') && ~isempty(e)
 end
 end
 
-function [fac, is_poisson, ymax] = parse_args(y, e, fac, is_poisson, use_local_max, varargin)
+function [fac, is_poisson, ymax] = parse_args(y, e, fac, is_poisson, varargin)
     p = inputParser;
     addRequired(p, 'y', @isnumeric);   % y compulsory
     addRequired(p, 'e', @isnumeric);   % e compulsory if any of remaining optional/parameter arguments present
     numeric_or_poisson = @(x) isnumeric(x) || strcmpi(x,'poisson');
     addOptional(p, 'dist_or_factor', fac, numeric_or_poisson);  % fac (numeric) or distribution ('poisson') optional, default to fac=0.1
-    addParameter(p,'maximum_value', use_local_max, @isnumeric);  % ymax, as parameter 'maximum_value'. Default to highly visible named negative value.
+    addParameter(p,'maximum_value', get_maximum_signal_value('default'), @isnumeric);  % ymax, as parameter 'maximum_value'. Default to internally set negative value.
     parse(p,y,e,varargin{:});
 
     % pick up signal max value as either default or input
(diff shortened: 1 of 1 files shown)


herbert_core/utilities/maths/noisify.m
```diff
@@ -25,32 +34,44 @@
...
+        ymax = max(abs(y(:)));          % find maximum magnitude of y for arbitrary dimensions
     end
-    ymax = max(abs(y(:)));          % find maximum magnitude of y for arbitrary dimensions
     dy=(fac*ymax)*randn(size(y));   % st. dev. of randn is sigma=1
     yout=y+dy;
     eout=ones(size(y))*(fac*ymax)^2;
 end
 
-if exist('e','var') && isempty(e)
+% this code now outputs a variance to be consistent between the use of eout
```
nickbattam-tessella, line 65: I don't think these two comments are adding any value for future developers:
- to anyone reading the code as it is now you have defined e as a variance at the top; so they wouldn't expect or think it was an stddev
- the missing `~` was a bug you've fixed. We don't need to document that beyond the git log

nickbattam-tessella (commented):
.

herbert_core/utilities/maths/noisify.m
```diff
@@ -25,32 +34,44 @@
...
             yout(i)=randpoisson(abs(y(i)));
         end
-        eout=abs(y);  % the inpuy y is the mean and variance of the Poisson distribution
+        eout=abs(y);  % the input y is the mean and variance of the Poisson distribution
     else
         error('Unrecognised random sampling distribution')
     end
-else
-    if nargin==3
+else % 3rd arg did not exist or was not a distribution name
```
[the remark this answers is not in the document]
hsautessella: You'll want to use [Matlab's `inputParser`](https://www.mathworks.com/help/matlab/ref/inputparser.html) if you're adding named arguments

cmarooney-stfc:
Recommitted to avoid defaulting fac with maxval present.

NB did not update header comments for this, more later.

cmarooney-stfc:
Recommitted to avoid defaulting fac with maxval present.

This commit does not use the parse_args method as I could not see how to work this with fac optionally not present.

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: Cleaned up with commit 4

herbert_core/utilities/maths/noisify.m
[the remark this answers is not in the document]
cmarooney-stfc: Fac made optional, 'maxval' flag introduced with commit 3. Not using input parser as optional fac gets in the way - for the one flag, alternative method used.

herbert_core/utilities/maths/noisify.m
```diff
@@ -17,40 +17,69 @@
...
+%           Omitting factor will use the default value 0.1.
+%
+%   If no input errors, e, just set e=[]. Note that eout will be created
+%   regardless of whether e is present, so the first overload output is
+%   probably incomplete.
+
+% deal with optional arguments
+is_poisson= false;
+fac = 0.1;
+if nargin>=2
```
hsautessella, line 36: Is this `if` statement necessary? Do you not get the same behaviour by removing this `if` block and making the call on line 38 `addOptional`?

herbert_core/utilities/maths/noisify.m
```diff
@@ -17,40 +17,69 @@
...
+    p = inputParser;
+    addRequired(p,'y');   % y compulsory
+    addRequired(p,'e');   % e compulsory if any of remaining optional/parameter arguments present
+    numeric_or_string = @(x) isnumeric(x) || ischar(x);
+    addOptional(p,'dist_or_factor',fac, numeric_or_string);  % fac (numeric) or distribution ('poisson') optional, default to fac=0.1
```
hsautessella, line 41: I'd put this block in a function at the bottom of the file e,g:
```matlab
function [y, e, dist_or_factor, maximum_value] = parse_args(y, e, varargin)
   ...
end
```

herbert_core/utilities/maths/noisify.m
```diff
@@ -17,40 +17,69 @@
...
+    addParameter(p,'maximum_value',-999,@isnumeric);  % ymax, as parameter 'maximum_value'. Default to highly visible negative value.
```
hsautessella, line 42: Could the default here be `max(y)` rather than -999?

herbert_core/utilities/maths/noisify.m
```diff
@@ -17,40 +17,69 @@
...
+        error('3rd argument cannot be interpreted as a Gaussian factor or legal probability distribution')
     end
+end
+
+% Use Poisson distribution and ignore other arguments
+if is_poisson
+    yout=zeros(size(y));
+    for i=1:numel(y)
+        yout(i)=randpoisson(abs(y(i)));
+    end
```
hsautessella, line 63: `y` could be a large array. Can this operation be vectorized? E.g.
```matlab
yout = randpoisson(abs(y))
```

herbert_core/utilities/maths/noisify.m
```diff
@@ -17,40 +17,69 @@
...
-    if is_stringmatchi(varargin{1},'poisson')
-        yout=zeros(size(y));
-        for i=1:numel(y)
-            yout(i)=randpoisson(abs(y(i)));
-        end
-        eout=abs(y);  % the inpuy y is the mean and variance of the Poisson distribution
+    % vary if 'poisson' or fac present
+    if isnumeric(p.Results.dist_or_factor)
+        fac = p.Results.dist_or_factor;
+    elseif p.Results.dist_or_factor == 'poisson'
```
hsautessella, line 51: Matlab has a [`strcmp`](https://www.mathworks.com/help/matlab/ref/strcmp.html) function that's usually used to compare strings. We may even consider `strcmpi` so the comparison is not case-sensitive

herbert_core/utilities/maths/noisify.m
```diff
@@ -17,40 +17,69 @@
...
+%   regardless of whether e is present, so the first overload output is
+%   probably incomplete.
+
+% deal with optional arguments
+is_poisson= false;
+fac = 0.1;
+if nargin>=2
+    p = inputParser;
+    addRequired(p,'y');   % y compulsory
+    addRequired(p,'e');   % e compulsory if any of remaining optional/parameter arguments present
```
hsautessella, line 39: Should we have a `@isnumeric` validator here?

hsautessella (changes requested):

herbert_core/utilities/maths/noisify.m
```diff
@@ -17,40 +17,69 @@
...
-        end
-        eout=abs(y);  % the inpuy y is the mean and variance of the Poisson distribution
+    % vary if 'poisson' or fac present
+    if isnumeric(p.Results.dist_or_factor)
+        fac = p.Results.dist_or_factor;
+    elseif p.Results.dist_or_factor == 'poisson'
+        is_poisson = true;
     else
-        error('Unrecognised random sampling distribution')
+        error('3rd argument cannot be interpreted as a Gaussian factor or legal probability distribution')
```
hsautessella, line 54: `error` also [takes an error ID argument](https://www.mathworks.com/help/matlab/ref/error.html#d122e350173). We've been trying to be more consistent in including identifiers in all our errors now. Having an identifier makes it easier to catch particular error types.

The identifier should have form `'CLASSNAME:method_name'`. Since this is a free function, I'd make the identifier `HERBERT:noisify`.

herbert_core/utilities/maths/noisify.m
nickbattam-tessella: If we can avoid introducing magic numbers it would be preferable.

herbert_core/utilities/maths/noisify.m
nickbattam-tessella: `e` is optional, but positionally significant. If you define it as required does the parser raise an error when it's missing?

cmarooney-stfc: e is required at that point in the code as it is within a nargin>=2 block, so it will never be missing as coded. Having e and disp_or_fac as both optional is problematic, hence the restriction. I tried isnumeric(y) in the debugger and it worked so I could add that.

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: We then have 2 optional parameters e and disp_or_fac with the possibility that e gets omitted and disp_or_fac does not. This is not a supported option from the header comments and I don't think I want to start choosing between optional parameters. If there are more than 2 parameters then e should be required. Ideally disp_or_fac should be a named parameter but that changes the existing interface.

herbert_core/utilities/maths/noisify.m
hsautessella: Why is it problematic if `e` and `disp_or_fac` are both optional?

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: If max(y) is done here then it will be wasted cpu time for the poisson case (also it is max(abs(y(:)))). I can name the parameter - -999 was a distinctive visual value, its only effect here is to be negative and not difficult to read like -1. What I can do is name the constant in addition so it is not so much a magic number.

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: There are also suggestions in the code that Toby would prefer e=[] rather than omitting it.

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: Guessing that since it is churning out random numbers, the generator is being advanced at each call. So not vectorizable. The equivalent built-in poissrnd does not seem to have this.

herbert_core/utilities/maths/noisify.m
hsautessella: The `addOptional` function adds a positional argument - so there's no way that one could specify `disp_or_fact` and not `e`.

herbert_core/utilities/maths/noisify.m
hsautessella: Good point - I hadn't spotted we weren't using it in the Poisson case. Would `-inf` be an even more obvious negative value?

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: This is not clear. All the inputParser docs say is that, combined with your/Nick's suggested code, is that e and disp_or_fac are both optional. I don't see how this form of argument parsing prevents leaving out e and using disp_or_fac - it will depend on the validation for e. It doesn't look like the normal method of arguments with defaults (ie optionals) where the first n optional arguments match.

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: That's a useful alternative. My measure of obvious/distinctive is "hits the eye of the beholder". Guessing that -999 is more "visual" than the string -inf - this is what you will see in the debugger. I was holding off using -666. But I would combine this with a named variable, I have coded USE_LOCAL_MAX to be visually distinctive in the code as distinct from the debugger.

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: As said above, you could leave out e and include disp_or_fac. I don't see anything in the inputParser docs which suggest this is impossible - it seems that ignoring it is done partly on the basis of the validation function argument, which will skip the argument slot if not validated and move on to the next one.

herbert_core/utilities/maths/noisify.m
hsautessella: Ah, you're right. I'd misunderstood the parser. I hadn't realised you could also specify parameters added via `addOptional` as named arguments.

herbert_core/utilities/maths/noisify.m
```diff
@@ -17,40 +17,75 @@
...
+    p = inputParser;
+    addRequired(p, 'y', @isnumeric);   % y compulsory
+    addRequired(p, 'e', @isnumeric);   % e compulsory if any of remaining optional/parameter arguments present
+    numeric_or_poisson = @(x) isnumeric(x) || strcmpi(x,'poisson');
+    addOptional(p, 'dist_or_factor', fac, numeric_or_poisson);  % fac (numeric) or distribution ('poisson') optional, default to fac=0.1
+    addParameter(p,'maximum_value', use_local_max, @isnumeric);  % ymax, as parameter 'maximum_value'. Default to highly visible named negative value.
+    parse(p,y,e,varargin{:});
+
+    % pick up signal max value as either default or input
+    ymax = p.Results.maximum_value;
```
nickbattam-tessella, line 81: Rather than pass a magic number as the maxval in as an argument, you could define that as a local (function level) magic number and move the    
```
if ymax == USE_LOCAL_MAX
        ymax = max(abs(y(:)));
end
```
block into this function.

Then the caller will have the correct value for maximum without needing any additional logic.

cmarooney-stfc: Pushed to do this. Now changed flag to -inf after consideration of existence of negative y values as per previous conversations.

herbert_core/utilities/maths/noisify.m
```diff
@@ -17,40 +17,79 @@
...
         yout=zeros(size(y));
         for i=1:numel(y)
             yout(i)=randpoisson(abs(y(i)));
         end
-        eout=abs(y);  % the inpuy y is the mean and variance of the Poisson distribution
+        eout=abs(y);  % the input y is the mean and variance of the Poisson distribution
     else
-        error('Unrecognised random sampling distribution')
+        if ~ymax_calculated
+            ymax = max(abs(y(:)))
```
nickbattam-tessella, line 51: this logic is repeated in the `parse_args` function. I see who it is needed here for `nargin == 1`. 

Can the `parse_args` function be adapted to handle that case too so the `ymax_calculated` flag and repeated logic are not required?

cmarooney-stfc: Undoubtedly, but this is working now, and the next work with noisify in Horace depends on it, so that improvement will happen later.

PR #239, opened 2020-09-25
Return RNG to previous state on create_folder cleanup

### LINKED ISSUE #238: RNG is being reset in try_to_create_folder
There's an `rng(...)` call in `try_to_create_folder` that is overwriting the RNG state. This is throwing off the random number generator in some of our tests. We should also not be overwriting users' RNG state.

The offending line is https://github.com/pace-neutrons/Herbert/blob/0906e1cfc3c8dff7668a4f64dd26e6275fea4b16/herbert_core/utilities/misc/try_to_create_folder.m#L32.

hsautessella:
Matlab's RNG state was being overwritten in `try_to_create_folder` which was throwing off some of our tests that use random number generation. We should also not be overwriting the RNG state of users.

I've added an `onCleanup` to reset the RNG's state.

Fixes #238
Fixes https://github.com/pace-neutrons/Horace/issues/385

### DIFF
--- a/_test/test_mpi/job_dispatcher_common_tests.m
+++ b/_test/test_mpi/job_dispatcher_common_tests.m
@@ -34,7 +34,7 @@ classdef job_dispatcher_common_tests < MPI_Test_Common
 
             % JETester specific control parameters
             old_rng_state = rng('shuffle');
-            cleanup = onCleanup(old_rng_state);
+            cleanup = onCleanup(@() rng(old_rng_state));
             FE = char(randi(25,1,5) + 64);
             common_param = struct('filepath', obj.working_dir, ...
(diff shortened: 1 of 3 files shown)


nickbattam-tessella:
The same call to `rng('shuffle')` is included in `Herbert\_test\test_mpi\job_dispatcher_common_tests.m` to generate random ids for inclusion in filenames.

Are the any assumptions made about the state of the random number generator in tests, other than the explicit seeding done for specific tests? 
Do these tests need to restore the state after being called?

hsautessella:
Any tests should be responsible for their own seeding - so those calls shouldn't affect other tests. That said, I think for sanity's sake, we should reset the RNG whenever we seed it. I'll update those. Good spot

nickbattam-tessella (approved):

PR #242, opened 2020-10-07
load phx fails when data are retrieved from cache

### LINKED ISSUE #241: load_nxspe does not work in phx mode
this issue identified from mslice and occurs if one tries to load parameters from nxspe file twice and request phx file format. 
On  the second attempt to load the data, the data, stored in memory as the structure, are not converted to array and conversion to phx fails.

abuts:
fixes #241 This small change fixes conversion to structure when data are taken from memory cache. The unit test shows the issue and the fixture in the code is trivial.


_test/test_data_loaders/test_nxspepar_loader.m
```diff
@@ -96,6 +97,21 @@ function test_mslice_par(this)
             [~,fname,fext] = fileparts(al.par_file_name);
             assertEqual([fname,fext],'MAP11014v2.nxspe');
         end
+        %
+        function test_mslice_par_cached(this)
```
hsautessella, line 101: Would `test_you_can_reload_a_cached_mslice_nxspe_par_file` be a more descriptive test name here?

hsautessella (approved):
Just a small comment on test naming, but not overly important.

Is the disabling of `build_sqw_in_parallel` related to this issue?

Otherwise, looks fine to me 👍

_test/test_data_loaders/test_nxspepar_loader.m
abuts: `build_sqw_in_parallel` is completely unrelated to this. 
as for name -- your suggestion is confusing. 
I would rather name it `mslice_par_from_cache_correct`  (when data are loaded from cache, they are still returned in correct format -- this was the issue in the first place).  It close to what I wrote, but if sounds better, I am changing it.

hsautessella (approved):
Thanks for the test rename. I think I was misunderstanding what it was doing, the new name + comment makes it clearer.

If the default config change is not related can we add something to the ticket description that explains/justifies this change? The commit says this was discussed in a meeting, but not how the decision to disable it was reached,

_test/test_data_loaders/test_nxspepar_loader.m
abuts: The first two commits of the PR come from some fixes done to release 3_4 done ages ago. We have not merged them into master directly, so I do not remember when it was done and what was discussed. Changes are simple and reasonable -- A user normally should not gen_sqw in parallel immediately -- I, for example, always do something wrong in couple of first runs.

abuts:
does my comment above counts for explanation?

hsautessella:
Yep, that's fine. Thank you

PR #244, opened 2020-10-14
Get Herbert noisify to choose what random number distribution it uses, provide test distribution for ease of debugging

cmarooney-stfc:
1) For herbert noisify, allow the user, to choose what random number distribution to use. This defaults to randn, the previous hard-coded choice, but enables testers to use other distributions which may be easier to test.

2) Add new (pseudo) random number distribution noisify_test_rand.m. This creates a regular ascending sequence .001, .002, .003 .... .998 .999 0 .001 .002 .... and so on. This enables testers to see manually by eye if the noise has been added correctly. It tracks adding this sequence across multiple calls, normally for separate pages of paged sqw signal data.


herbert_core/utilities/maths/noisify_test_rand.m
```diff
@@ -0,0 +1,53 @@
...
+% The global flag is expected to be acceptable in a test
+% though not in production code.
+%--------------------------------------------------
+
+% external global flag to signal that this pseudo-random
+% number distribution needs to be initialised if true.
+global noisify_test_rand_init;
+% internal number tracking where in the generation of
+% the sequence we have got to.
+persistent got_to;
```
nickbattam-tessella, line 27: is there any way to reset this?

If there are multiple tests using this test function the results will vary depending on whether coupled tests have been run

nickbattam-tessella (commented):
Since `noisify_test_rand` is a utility function that only has use inside the test framework it may fit better in `_test\shared` (with `seed_rng.m`) which is accessible from Horace and Herbert.

herbert_core/utilities/maths/noisify_test_rand.m
cmarooney-stfc: this function now recast as class object with latest commit

cmarooney-stfc: and moved to _test/shared

herbert_core/utilities/maths/noisify.m
```diff
@@ -51,7 +52,7 @@
             ymax = max(abs(y(:)))
         end
         % make noise dy and add to y for output; make error bar for noise    
-        dy=(fac*ymax)*randn(size(y));   % st. dev. of randn is sigma=1
+        dy=(fac*ymax)*randfunc(size(y));% default: randn(size(y)); % st. dev. of randn is sigma=1
```
nickbattam-tessella, line 55: Nested comments. Also the `default:` here just repeats the logic of the code in lines 37-41

_test/shared/noisify_test_rand.m
```diff
@@ -0,0 +1,65 @@
+classdef noisify_test_rand < handle
+%--------------------------------------------------
+% This class creates a simple deterministic series
+% of numbers which can be substituted for a true pseudo-random
+% number generator to enable by-eye checking of the random
+% noise addition (via noisify( to the signal. It can
```
nickbattam-tessella, line 6: `(via noisify(` < typo on closing bracket

_test/shared/noisify_test_rand.m
```diff
@@ -0,0 +1,65 @@
+classdef noisify_test_rand < handle
```
nickbattam-tessella, line 1: As this is a shared RNG class it is not inherently a "noisify" specific function so that shouldn't be part of the class name or be included in the function comment.

This is a pseudo random number generator which returns an ordered, linear  series, as described in the comment further down: 
```
%The values generated are an ascending sequence of increment one in 
% the range [0:999]*1e-3.
```

_test/shared/noisify_test_rand.m
```diff
@@ -0,0 +1,65 @@
...
+        function vals = myrand(obj,size)
+            % On initialisation the flag is reset false and the number
+            % sequence tracker is initialised to 1
+            if obj.init==1
+                obj.init=0;
+                obj.got_to=1;
+            end
+            
+            % the size vector of the signal which is passed in as the argument
+            % is a 1xsize vector - the actual size is extracted from the second
```
nickbattam-tessella, line 40: this comment is describing how MATLAB works.

_test/shared/noisify_test_rand.m
```diff
@@ -0,0 +1,65 @@
...
+                obj.init=0;
+                obj.got_to=1;
+            end
+            
+            % the size vector of the signal which is passed in as the argument
+            % is a 1xsize vector - the actual size is extracted from the second 
+            % element.
+            size=size(2);
+            
+            % The values generated are an ascending sequence of increment one in
```
nickbattam-tessella, line 44: this comment described what the function does and expected input so belongs in the function header rather than the code.

nickbattam-tessella (changes requested):
The fake-rng function is not used anywhere, nor are there any tests for it.

_test/shared/noisify_test_rand.m
cmarooney-stfc: No, it's describing how the expected argument will function. It is possible for a user to input a simple scalar - which would be wrong - this interface is expecting size(y) where y is the signal vector.

_test/shared/noisify_test_rand.m
cmarooney-stfc: Sometimes it is useful for the maintainer to understand what is going on with the particular line of code. This is the implementation. If it went in the header it would be describing the output, the expected interface. It would also be good to indicate whether your comment is saying that something is wrong - doesn't do what is expected - vs does not adhere to some protocol.

_test/shared/noisify_test_rand.m
cmarooney-stfc: Well, you were the one who asked for it to go in the generic utilities location. Maybe it should go back to the noisify location.

_test/shared/noisify_test_rand.m
cmarooney-stfc: Yes

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: RH comment is the original comment on the randn call. LH comment is commenting on the use of randfunc. It seems worth ramming home that randn is the "normal" use-case, not just a standard option if the user can't think of anything more specific. It is always good to restate things in different ways to get the point across.

hsautessella (commented):
As a small style note: we seem to have inconsistency with spacing between operators. E.g. [line 31](https://github.com/pace-neutrons/Herbert/pull/244/files#diff-05368fa48da82c99c00e56cc5a57414420bdcfb814959ae96603cebac42d1f11R31) has spaces around the equals sign, whereas [lines 34-36](https://github.com/pace-neutrons/Herbert/pull/244/files#diff-05368fa48da82c99c00e56cc5a57414420bdcfb814959ae96603cebac42d1f11R34-R36) do not. 

I realise existing code in Horace is very inconsistent on this, but we've been preferring spaces around `=`.

_test/shared/noisify_test_rand.m
nickbattam-tessella: I was suggesting that it should be here, but should be renamed without 'noisify' in the name.

herbert_core/utilities/maths/noisify.m
nickbattam-tessella: Fair enough.
I would suggest pulling that whole inline comment out to the preceding line then:
```
% make noise ....
% default: normal distribution, stddev = 1 (randn(size(y)))
dy=(fac*ymax)*randfunc(size(y));
```
That contains the same information but means I don't have to remember what randn does and not having to parse nested comments
```
code; % comment1 % comment2
```

_test/shared/noisify_test_rand.m
nickbattam-tessella: It's a stylistic thing rather than a right/wrong thing.

The function name is a short summary of what the function does (get_x, add_values, ...) and argument names hints to what the expected input/output arguments should be.

The function header is where a user would look for information about input and output arguments and detail about what the function will do. There shouldn't be need for anyone to look at implementation of functions to understand how to call them or what the effect will be.

The code comment is, as you say, information for the developer on how that is achieved if it's not obvious from the code. or why it's implemented in that way.

_test/shared/noisify_test_rand.m
nickbattam-tessella: I understand.

This distinction would be clearer if rather than reassigning the `size` variable here a new variable, e.g. "num_elements" or "signal_length" was created.

_test/shared/noisify_test_rand.m
```diff
@@ -0,0 +1,65 @@
...
+    properties
+        % flag to signal that this pseudo-random
+        % number distribution needs to be initialised if true.
+        init = 1
+        % number tracking where in the generation of
+        % the sequence we have got to. Initially a dummy bad value.
+        got_to = -666
+    end
+    methods
+        function vals = myrand(obj,size)
```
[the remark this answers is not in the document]
cmarooney-stfc: Both excellent comments. Doubly needed as I am struggling with a stiff spacebar on the other laptop - I may have intended what you suggest

_test/shared/noisify_test_rand.m
cmarooney-stfc: My favourite book is Conway's "Perl best practices". A lot of it could be summarized as "actually use Python", but apart from that... it has useful comments like "you can  restrict doing [elaborate comment method] to code of more than N lines long .... as long as N=0". Here "if it's not obvious" can be expanded with "(and it will be not obvious)". Developers often tend to start reading the code directly, and only give up and look at the header if it becomes too boggy. Especially true if you are in the debugger where you do start at the first executable line.

_test/shared/noisify_test_rand.m
cmarooney-stfc: That is useful. Harry comments that size being a matlab function it should not be reused as a variable name - fixing that will cover this.

_test/shared/noisify_test_rand.m
cmarooney-stfc: Suggest you check your original comment

herbert_core/utilities/maths/noisify.m
```diff
@@ -50,8 +56,13 @@
...
             ymax = max(abs(y(:)))
         end
-        % make noise dy and add to y for output; make error bar for noise    
-        dy=(fac*ymax)*randn(size(y));   % st. dev. of randn is sigma=1
+        % make noise dy and add to y for output; make error bar for noise.  
+        % randfunc generates the random numbers producing the noise;
+        % its standard and default implementation is randn - gaussian of std.dev=1
+        % but other choices may be appropriate including a non-random
+        % sequence to check functionality. Not using randn may cause
+        % unexpected behaviour from the user point of view
```
hsautessella, line 64:
```suggestion
        % unexpected behaviour from the user's point of view
```

_test/test_utilities/test_deterministic_pseudorandom_sequence.m
```diff
@@ -0,0 +1,68 @@
...
+        % TestClass instance and minimally displays the name argument.
+        function this=test_deterministic_pseudorandom_sequence(name)
+            this = this@TestCase(name);
+            disp(name)
+        end
+        
+        % test function to verify operation of noisify_test_rand
+        function test_if_deterministic_pseudorandom_sequence_is_ok(this)
+            % find noisify_test_rand
+            addpath(this.deterministic_pseudorandom_sequence_path);
```
hsautessella, line 30: If this test fails then the `rmpath` call on line 65 won't be reached and the path added here will stay on the path. The solution to this is to use a [cleanup object](https://www.mathworks.com/help/matlab/ref/oncleanup.html). 

E.g. 
```matlab
addpath(this.deterministic_pseudorandom_sequence_path);
cleanup = onCleanup(@() rmpath(this.deterministic_pseudorandom_sequence_path);
```

_test/test_utilities/test_deterministic_pseudorandom_sequence.m
```diff
@@ -0,0 +1,68 @@
...
+            my_sequence_gen = deterministic_pseudorandom_sequence();
+            % create one sequence of size sy1
+            % NB sizes are designed to push the sequence through 
+            %    the value 999 where the sequence restarts from 0
+            sy1 = 1191;
+            dy1 = my_sequence_gen.myrand([1,sy1]);
+            % create a second sequence of the same size, starting
+            % after the endpoint of the first sequence
+            sy2 = 1191;
+            dy2 = my_sequence_gen.myrand([1,sy2]);
```
hsautessella, line 43: This test seems to be testing several things. 

It looks like you're testing that:

1. Two sequences of the same size are continued on consecutive calls
2. The sequence generator can be reset
3. Two sequences of different sizes are continued on consecutive calls

I'd have these as separate tests if that's possible.

herbert_core/utilities/maths/noisify.m
```diff
@@ -67,15 +78,21 @@
...
     addOptional(p, 'dist_or_factor', fac, numeric_or_poisson);  % fac (numeric) or distribution ('poisson') optional, default to fac=0.1
     addParameter(p,'maximum_value', USE_LOCAL_MAX, @isnumeric);  % ymax, as parameter 'maximum_value'. Default to internally set negative value.
-    parse(p,y,e,varargin{:});
+    check_function_handle = @(x) isa(x,'function_handle');
+    addParameter(p,'random_number_function', @randn, check_function_handle); 
+    try
+        parse(p,y,e,varargin{:});
+    catch ME
+        disp(ME);
+    end
```
hsautessella, line 95: It's not immediately clear why we're catching and displaying this error but not exiting. `ME` is a struct, did you want `ME.message`?

hsautessella (commented):
This looks good to me. Just a couple of minor things

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: A good catch. Thanks for the info on ME.message. Should probably be removed; I was trying to track down a failure.

_test/test_utilities/test_deterministic_pseudorandom_sequence.m
cmarooney-stfc: This is useful, thanks. But in what sense does it remain on the path? When the test concludes, what environment has this path? I was expecting that the environment would be destroyed at end of test.

herbert_core/utilities/maths/noisify.m
cmarooney-stfc: Your correction is acceptable, but using "user",  as an adjective qualifying point, seems perfectly good to me. Compare "user experience" vs "user's experience".

_test/test_utilities/test_deterministic_pseudorandom_sequence.m
hsautessella: Unless I'm mistaken, it's added to the global path for that Matlab session. It's not just added for the scope of the test - you need to manually remove it.

_test/test_utilities/test_deterministic_pseudorandom_sequence.m
```diff
@@ -0,0 +1,192 @@
...
+            % restore original path
+            %moved tg onCleanUp: rmpath(this.deterministic_pseudorandom_sequence_path);
+        end
+        
+        % test function to verify operation of noisify_test_rand
+        % tests repeated usage for one distribution of equal size
+        function test_if_equal_twocall_deterministic_pseudorandom_sequence_ok(this)
+            % find noisify_test_rand
+            addpath(this.deterministic_pseudorandom_sequence_path);
+            cleanup = onCleanup( @() rmpath(this.deterministic_pseudorandom_sequence_path) );
```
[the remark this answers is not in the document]
nickbattam-tessella: The test framework we're using supports both `setUp`/`tearDown` methods that are run before/after each test, and class level setup and teardown code which can be put in the class constructor  and `delete` methods. 

There are some examples of this is use e.g. `Horace\_test\test_gen_sqw_workflow\gen_sqw_common_config.m`

nickbattam-tessella (approved):

PR #249, opened 2020-11-05
Manage sqw class rename in test data snapshot files

nickbattam-tessella:
Add preprocess shim to manage conversion of sqw to sqw_old in test snapshot files; temporary code that will be removed.

### DIFF
--- a/_test/shared/matlab_xunit_ISISextras/manage_legacy_sqw_class_rename.m
+++ b/_test/shared/matlab_xunit_ISISextras/manage_legacy_sqw_class_rename.m
@@ -12,10 +12,16 @@ elseif isstruct(data)
     field_names = fields(data);
     for idx = 1:length(field_names)
         field_name = field_names{idx};
-        for inner_idx = 1:numel(data.(field_name))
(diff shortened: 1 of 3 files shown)


_test/shared/matlab_xunit_ISISextras/manage_legacy_sqw_class_rename.m
```diff
@@ -0,0 +1,23 @@
+function processed = manage_legacy_sqw_class_rename(data)
+% Perform pre-processing on data loaded from .mat files as part of the
+% TestCaseWithSave workflow.
+%
+% This is a temporary function to manage reengineering of the `sqw` object
+% during which the old class is renamed as `sqw_old`, and will load as
+% a struct rather than class instance.
+
+    if isfield(data, 'main_header') && isfield(data, 'header') && isfield(data, 'detpar')
+        processed = sqw_old(data);
```
hsautessella, line 10: The function header isn't indented but the code is. I usually do not indent in top level functions - may as well save the horizontal whitespace.

hsautessella (approved):
Looks good to me - I can't really manually test without having the `sqw_old` class, but it looks like it should work!

One very minor comment that I wouldn't worry about unless you're making any other changes.

hsautessella (approved):

PR #251, opened 2020-11-06
bug if worker missing

abuts:
This is the test for the issue, identified if worker is missing and some users try to set up outdated configuration


nickbattam-tessella (approved):
All references to 'parallel_framework' removed from code.

PR #260, opened 2020-11-22
fix TestCaseWithSave for parent class

### LINKED ISSUE #259: TestWithSave does not work for common parent class
If `assertEqualWithSave(this,test_obj)` is invoked from common test class (e.g. in Horace, `test_gen_sqw_accumulate_sqw_mex` and `test_gen_sqw_accumulate_sqw_nomex`  are testing the same code with different parameters, so both are the children of `gen_sqw_accumulate_sqw_common` class and the constructor of the particular classes just sets appropriate parameters, the method does not recognizes the class name and fails. 

This is simple bug which should be fixed.

abuts:
fixes #259

### DIFF
--- a/_test/shared/matlab_xunit_ISISextras/@TestCaseWithSave/TestCaseWithSave.m
+++ b/_test/shared/matlab_xunit_ISISextras/@TestCaseWithSave/TestCaseWithSave.m
@@ -641,6 +641,8 @@ classdef TestCaseWithSave < TestCase & oldTestCaseWithSaveInterface
                         strncmpi(cont{end},'test',4)
                     test_name = cont{end};
                     break
+                else
+                    test_name = cont{end};
                 end
             end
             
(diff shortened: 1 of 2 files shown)


_test/shared/matlab_xunit_ISISextras/@TestCaseWithSave/TestCaseWithSave.m
```diff
@@ -641,6 +641,8 @@ function assertMethodWithSave (this, var, var_name, funcHandle, varargin)
                         strncmpi(cont{end},'test',4)
                     test_name = cont{end};
                     break
+                else
+                    test_name = cont{end};
```
nickbattam-tessella, line 645: The `if` and `else` block perform the same action. Is that check required?

nickbattam-tessella: Looking again the `break` is different in two branches, so it can be rewritten with just that in the `if`
```
test_name = cont{end};
if strcmp && ~strcmp && strncmpi
  break;
end
```

abuts: yes -- this is better

PR #265, opened 2020-11-25
truncate asciipar

abuts:
Better and more consistent implementation of the asciipar truncation with the purpose of producing consistent results on any OS with any code


nickbattam-tessella (commented):
Would there be value in declaring the `4` in load_ASCII_par as a constant. It is used repeatedly and if ever changed again having one place to change it would be simpler.

abuts:
fair pint

abuts:
> 
> 
> Would there be value in declaring the `4` in load_ASCII_par as a constant. It is used repeatedly and if ever changed again having one place to change it would be simpler.

fair point

Issue #269, opened 2020-11-30
changes, related to urange clarification


abuts:
This is ticket to include appropriate changes for Horace ticket [#447](https://github.com/pace-neutrons/Horace/issues/447), related to urange clarification

cmarooney-stfc:
@abuts Hi Alex, checking that this Herbert ticket is still blocking Horace #447?

abuts:
well, these changes are necessary for Horace#447. It is blocking for it. I will merge it as soon as #447 is completed. Some tests in #447 are still failing and though I hope no changes here are necessary to fix it, I keep it open until all changes in #447 are done.

cmarooney-stfc:
@abuts Hi Alex, it looks as if this is finished as the associated Horace #447 is now closed. Can we close this one please?

PR #271, opened 2020-12-04
Const bin boundaries and unit test for it

abuts:
changes from Re #269  This is simple addition of the code, responsible for axis binning. Related to uranges and generic projections

### DIFF
--- a/_test/test_utilities/test_make_const_bin_boundaries.m
+++ b/_test/test_utilities/test_make_const_bin_boundaries.m
@@ -57,6 +57,7 @@ classdef test_make_const_bin_boundaries < TestCase
             assertEqual(numel(p),11);
             assertEqual(p',0.5:1:10.5)
             
+            % p_des should be equal to pbin_initial (to within rounding errors)
             p_des = make_const_bin_boundaries_descr(p);
             assertEqual(p_des,pbin);
         end
--- a/_test/test_utilities/test_make_const_bin_boundaries.m
+++ b/_test/test_utilities/test_make_const_bin_boundaries.m
@@ -1,0 +1,64 @@
+classdef test_make_const_bin_boundaries < TestCase
+    
+    
+    properties
+    end
+    %
+    methods
+        function obj = test_make_const_bin_boundaries(varargin)
+            if nargin>0
+                name = varargin{1};
+            else
+                name = 'test_make_const_bin_boundaries';
+            end
+            obj = obj@TestCase(name);
+        end
+        function test_boundaries_with_zero(~)
+            pbin = [1,0,10];
+            range = [2,9];
+            pref = 1;
+            p=make_const_bin_boundaries(pbin,range,pref);
+            assertEqual(numel(p),11);
+            assertEqual(p',0.5:1:10.5);
+            
+            pbin = [-inf,0,10];
+            p=make_const_bin_boundaries(pbin,range,pref);
+            assertEqual(numel(p),10);
+            assertEqual(p',1.5:1:10.5);
+            
+            pbin = [-inf,0,inf];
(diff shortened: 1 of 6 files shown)


herbert_core/utilities/misc/make_const_bin_boundaries.m
```diff
@@ -0,0 +1,188 @@
...
+%
+% Ouput:
+% ------
+%   p       Column vector of bin boundaries
+
+
+% Check input arguments
+% -----------------------
+if ~isnumeric(pbin) || numel(pbin)~=3
+    error('Binning argument is invalid data type or length')
```
nickbattam-tessella, line 49: Should all the errors thrown from this function include an errID (e.g. `HERBERT:UTILITIES:MAKE_CONST_BIN_BOUNDARIES`)?

abuts: fair point

_test/test_utilities/test_make_const_bin_boundaries.m
```diff
@@ -0,0 +1,64 @@
+classdef test_make_const_bin_boundaries < TestCase
+    
+    
+    properties
+    end
+    %
+    methods
+        function obj = test_make_const_bin_boundaries(varargin)
+            if nargin>0
```
nickbattam-tessella, line 9: why do you support the test name overloading here rather than simply call `obj = obj@TestCase('test_make_const_bin_boundaries';);`

herbert_core/utilities/misc/make_const_bin_boundaries_descr.m
```diff
@@ -0,0 +1,25 @@
...
+%   >> pbin = make_const_bin_boundaries_descr_(p)
+%
+% Input:
+% ------
+%   p       Bin boundaries, assumed constant spacing
+%
+% Output:
+% -------
+%   pbin    Bin boundary descriptor [pbeg, pstep, pend] where
+%          pbeg in the forst bin centre, pstep is the bin width, and
```
nickbattam-tessella, line 13: typo: `first`

herbert_core/utilities/misc/make_const_bin_boundaries_descr.m
```diff
@@ -0,0 +1,25 @@
+function pbin = make_const_bin_boundaries_descr(p)
+% Make a constant bin boundaries descriptor from the bin boundaries
```
nickbattam-tessella, line 2: Maybe worth extending this comment to make it clear the way this works with `make_const_bin_boundaries`

```
pbin_initial = [1, 1, 10];
p=make_const_bin_boundaries(pbin_initial );
p_desc = make_const_bin_boundaries_descr(p);

% p_desc is equal to pbin_initial (to within rounding errors)
```

_test/test_utilities/test_make_const_bin_boundaries.m
abuts: The original Tests were creating test object for every tests method. We have rewritten that to have one constructor per test class. 
When the constructor was called per method, single test class name would not work. But when you run runtests, it still names test classes according to its own naming. I have never cared to investigate is it necessary indeed or one may call the classes as internal name.  Its clear, varargin is needed for a test input as a default name is provided for tests. And if you have varargin then why not use it. 

The default name is convenient when you run single tests in a pattern:
tc = test_specific_tests
tc.test_method();
then default name is needed. 

There is also minor convenience -- possibility to have the same name for different test classes. 

So, in a way, it is not really necessary any more but may be convenient and useful in special cases and the custom remain from the time it was necessary

abuts:
not so important as descriped in the functions but makes no harm

PR #280, opened 2021-01-21
528 replace disabled skip

oerc0122:
Replace Herbert's disabled/skipped tests with proper skip


_test/test_mpi/test_ParpoolMPI_Framework.m
```diff
@@ -53,41 +53,36 @@
...
+
         end
         function delete(obj)
             obj.pool_deleter = [];
         end
         %
         function test_finish_tasks_reduce_messages(obj,varargin)
             if obj.ignore_test
-                return;
+                skipTest();
```
hsautessella, line 64: Might we put "_Parallel computing toolbox is not available_" as the reason for skipping this, and the following, tests?

_test/test_mpi_wrappers/test_exchange_CppMPI.m
```diff
@@ -19,47 +19,47 @@
         function test_OutOfRange(obj)
             % Test communications in test mode
             if obj.ignore_test
-                return
+                skipTest();
```
hsautessella, line 22: Quite a few of these tests are skipped on whether `MPI_Test_Common.ignore_tests`  is true. Might it be worth adding a `.skip_reason` attribute to `MPI_Test_Common` that records why the test was skipped?

hsautessella (commented):
This looks good, just one or two tests where a skip reason could be added

_test/test_mpi/test_ParpoolMPI_Framework.m
```diff
@@ -97,34 +92,30 @@ function test_finish_tasks_reduce_messages(obj,varargin)
...
             assertEqual(mess.mess_name,'completed');
-            
+
         end
         %
         function test_probe_all_receive_all(obj,varargin)
             % common code -------------------------------------------------
             if obj.ignore_test
-                return;
+                skipTest();
```
cmarooney-stfc, line 101: If this code is now registering something - a skip - rather than silently existing, worth providing a reason argument so people seeing the skip can know to ignore it

_test/test_mpi/test_ParpoolMPI_Framework.m
```diff
@@ -152,38 +143,35 @@ function test_probe_all_receive_all(obj,varargin)
...
             end
-            
+
         end
         %
         function test_send_receive_message(obj,varargin)
             % common code -------------------------------------------------
             if obj.ignore_test
-                return;
+                skipTest();
```
cmarooney-stfc, line 152: ditto

_test/test_mpi/test_job_dispatcher_parpool.m
```diff
@@ -17,20 +17,20 @@ function delete(~)
...
         %
-        
+
         function xest_job_submittion(obj)
             % test to debug job submission on cluster. It's not usually run
             % as all logic is tested elsewhere but kept to help identifying
             % the issues with job submission on a cluster.
             if obj.ignore_test
-                return;
+                skipTest();
```
cmarooney-stfc, line 26: ditto

_test/test_utilities/test_get_revert_bytestream.m
```diff
@@ -12,72 +12,71 @@
...
-            
+
         end
-        
+
         function test_mex_nomex(this)
-            if matlab_version_num>8.02 % Matlab 2014a can not currently run mex files build for lower matlab versions
-                return
+            if matlab_version_num>8.02
+                skipTest('Matlab 2014a can not currently run mex files build for lower matlab versions');
```
cmarooney-stfc, line 61: I appreciate that this is an existing test, rather than something you have created, but the test here checks that matlab is 2014a or later (so worth adding "and greater") but does not appear to have a test for the mex file versions.

_test/test_utilities/test_resolve_path_userpath.m
```diff
@@ -11,91 +11,91 @@
...
-            
+
             assertEqual(actual, one_up);
-            
+
         end
         function test_resolve_home_dir(~)
             if ~isunix
-                return;
+                skipTest('No "~" home directory on Windows')
```
cmarooney-stfc, line 96: It looks as if the test is just to be skipped on windows or mac. I can't see anything about home directory - the ~ in the argument is not the same (I think) as the ~ in the message.

cmarooney-stfc (commented):
Looking at the large amount of whitespace changes, this creates a lot of noise and makes it more difficult to review. Is this due to automatic reformatting?

_test/test_mpi/test_ParpoolMPI_Framework.m
oerc0122: It defaults to "skipped test". The warning is provided higher up when ignore test is set.

_test/test_utilities/test_resolve_path_userpath.m
oerc0122: Two lines down:
```
            expected = getuserdir();
            actual = resolve_path('~');
```
Mac does have a `~` directory, hence.

oerc0122:
> Looking at the large amount of whitespace changes, this creates a lot of noise and makes it more difficult to review. Is this due to automatic reformatting?

@cmarooney-stfc Click the gear icon in the comparison (top left), `Hide whitespace changes`.

_test/test_utilities/test_resolve_path_userpath.m
nickbattam-tessella: I am enjoying the three different meanings of the `~` symbol on three successive lines.

nickbattam-tessella: Is there a windows equivalent of this test which checks the Windows use-case against `%USERPROFILE` (outside the scope of this PR)

PR #281, opened 2021-01-26
Masking disabled on keys

abuts:
This small change is related to Horace ticket [#464](https://github.com/pace-neutrons/Horace/issues/464) and enables masking according to keys, e.g. disabling pixels with signal equal to nan or inf

### DIFF
--- a/_test/test_data_loaders/test_rundata_rm_masked.m
+++ b/_test/test_data_loaders/test_rundata_rm_masked.m
@@ -34,7 +34,7 @@ classdef test_rundata_rm_masked< TestCase
             assertEqual(run.ERR,err);
             assertEqual(run.det_par,det);
         end
-        function test_works_removesNAN(~)
+        function test_works_removesNaNandInf(~)
             run=rundata();
             run.S=ones(3,5);
             run.ERR=ones(3,5);
@@ -67,7 +67,7 @@ classdef test_rundata_rm_masked< TestCase
             assertEqual(numel(det.width),5);
         end
         
-        function test_inf_masking_disabled(~)
+        function test_Inf_masking_disabled(~)
             run=rundata();
             run.S=ones(3,5);
             run.ERR=ones(3,5);
@@ -84,7 +84,7 @@ classdef test_rundata_rm_masked< TestCase
             assertEqual(numel(det.width),4);
         end
         
-        function test_nan_masking_disabled(~)
+        function test_NaN_masking_disabled(~)
             run=rundata();
             run.S=ones(3,5);
             run.ERR=ones(3,5);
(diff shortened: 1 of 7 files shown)


herbert_core/classes/data_loaders/@rundata/rm_masked.m
```diff
@@ -1,19 +1,37 @@
...
+if ~exist('ignore_inf','var')
+    ignore_inf = true;
+end
 
-index_masked = (isnan(this.S)|(isinf(this.S))); % masked pixels
+
+if ignore_nan && ignore_inf
+    index_masked = (isnan(obj.S)|(isinf(obj.S))); % masked pixels
+elseif ignore_nan
+    index_masked = (isnan(obj.S));
```
nickbattam-tessella, line 22: Are the outer brackets around the operations in these three cases required? They make the code harder to read

_test/test_data_loaders/test_rundata_rm_masked.m
```diff
@@ -35,22 +34,74 @@ function test_works_do_nothing(this)
             assertEqual(run.ERR,err);
             assertEqual(run.det_par,det);
         end
-        function test_works_removesNAN(this)
+        function test_works_removesNAN(~)
```
nickbattam-tessella, line 37: Test should be renamed to say it removes INF and NaN by default, and an additional test created confirming that both are removed if `rm_masked(run, true, true);` is called would fully demonstrate the API

nickbattam-tessella (changes requested):
Is this code used anywhere except in the test class you've updated?

herbert_core/classes/data_loaders/@rundata/rm_masked.m
```diff
@@ -1,30 +1,58 @@
...
-index_masked = (isnan(this.S)|(isinf(this.S))); % masked pixels
-line_notmasked= ~any(index_masked,1);           % masked detectors (for any energy)
+if ignore_nan && ignore_inf
+    index_masked = isnan(obj.S)| isinf(obj.S); % masked pixels
+elseif ignore_nan
+    index_masked = isnan(obj.S);
+elseif ignore_inf
+    index_masked = isinf(obj.S);
+else
+    S_m= obj.S;
```
nickbattam-tessella, line 34: It would be clearer to move the 'early return' logic into a separate if block before the `index_masked` logic, i.e.
```
if ~ignore_nan && ~ignore_inf
  ...set stuff...
  return
end

if ignore_nan && ignore_inf
   index_masked = isnan(obj.S)| isinf(obj.S); % masked pixels
   ...
```

It's easy to miss a `return` buried in a multi-branch if.

abuts: probably. Difficult to judge. Will try to change -- it leaves one of the if branches hanging, which is not obvious too.

nickbattam-tessella (approved):

PR #291, opened 2021-02-08
Add MLint warnings to build

oerc0122:
Run MLint on build (Horace #474)
Update warning to use Warnings Next Gen (Horace #547)


CMakeLists.txt
```diff
@@ -18,12 +18,14 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake"
...
 # This sets the destination for mex build artifacts (used in PACE_AddMex)
 set(Herbert_DLL_DIRECTORY "${CMAKE_SOURCE_DIR}/herbert_core/DLL")
+set(Herbert_ROOT "${Herbert_SOURCE_DIR}/herbert_core")
 
 # Look for packages early so we can exit if they're not found
 include(herbert_FindMPI)
 include(PACE_FindMatlab)
 include(PACE_Version)
 include(PACE_AddMex)
+include(CodeAnalysis)
```
nickbattam-tessella, line 28: does this always need including?

admin/lint_json.m
```diff
@@ -0,0 +1,52 @@
+function lint_json(filesin, outputfile)
```
nickbattam-tessella, line 1: This would benefit from a header comment as it's not clear what it's doing from the name

tools/build_config/build.sh
```diff
@@ -71,18 +71,10 @@ function run_tests() {
...
+  local build_dir=$1
 
-    analysis_cmd="cppcheck --enable=all --inconclusive"
-    analysis_cmd+=" --xml --xml-version=2"
-    analysis_cmd+=" -I ${HERBERT_ROOT}/_LowLevelCode/cpp"
-    analysis_cmd+=" ${HERBERT_ROOT}/_LowLevelCode/"
-    analysis_cmd+=" 2> ${output_dir}/cppcheck.xml"
-    echo_and_run "${analysis_cmd}"
-  fi
+  echo_and_run "cd ${build_dir}"
```
nickbattam-tessella, line 76: This pair can be rewritten as `run_in_dir`

cmake/shared/CodeAnalysis.cmake
```diff
@@ -0,0 +1,24 @@
+add_custom_target(analyse
+  COMMENT "Performing code analysis..."
+  )
+
+add_custom_target(analyse-mlint
+  COMMENT "- Performing MATLAB analysis (Mlint)..."
+  BYPRODUCTS "${CMAKE_CURRENT_BINARY_DIR}/mlint.out"
+  COMMAND matlab -nodisplay <<< "\"addpath('${Herbert_ROOT}/../admin');lint_json({'${CMAKE_SOURCE_DIR}/**/*.m'},'${CMAKE_CURRENT_BINARY_DIR}/mlint.json');\""
```
nickbattam-tessella, line 8: The CmakeList.txt in _test folder include code managing the MATLAB path before executing tests

CMakeLists.txt
oerc0122: Doesn't hurt to do so and allows users to run manual analysis if they so wish. It adds the option to do code analysis, not the requirement.

tools/build_config/build.sh
```diff
@@ -178,15 +174,18 @@ function main() {
...
         This may not be a clean build."
     echo_and_run "mkdir ${build_dir}" || warning "${warning_msg}"
     run_configure "${build_dir}" "${build_config}" "${build_tests}" "${matlab_release}" "${cmake_flags}"
+  fi
+
+  if ((analyze)); then
+    run_analysis "${build_dir}"
+  fi
+
+  if ((build)); then
```
hsautessella, line 179: To avoid us having to add the `--configure` flag when we want to build (which would make this script out-of-sync with the pwsh equivalent), would it be possible to add an `if ... fi` here and call `run_configure` if it's not been run yet? Checking for the existence of `${build_dir}/CMakeCache.txt` should be a sufficient.

tools/build_config/Jenkinsfile
```diff
@@ -294,14 +293,13 @@ pipeline {
...
         fingerprint: true
       )
 
-      publishCppcheck (
-        displayAllErrors: true,
-        allowNoReport: true,
-        pattern: '**/cppcheck.xml'
+      recordIssues(tools:
+                   [cppCheck(pattern: '**/cppcheck.xml', reportEncoding: 'UTF-8'),
+                    issues(id: 'mlint', name: 'MLint', pattern: '**/mlint.json')]
```
hsautessella, line 302:
```suggestion
      recordIssues(
        tools: [
          cppCheck(pattern: '**/cppcheck.xml', reportEncoding: 'UTF-8'),
          issues(id: 'mlint', name: 'MLint', pattern: '**/mlint.json')
        ]
```

would be more consistent with the surrounding formatting

admin/lint_json.m
```diff
@@ -0,0 +1,52 @@
...
+    raw = checkcode(files, '-id');
+    for i = 1:numel(raw)
+        for j = 1:numel(raw{i})
+            raw{i}(j).fileName = files{i};
+            curr = wng_compat(raw{i}(j));
+            issues.issues = {issues.issues{:}, curr};
+        end
+    end
+    issues.size = numel(issues.issues);
+    fprintf(fh, "%s", jsonencode(issues));
```
hsautessella, line 42: `fprintf` has a limited buffer size, `fwrite` is probably the safer option here

admin/lint_json.m
```diff
@@ -0,0 +1,52 @@
+function lint_json(filesin, outputfile)
+    if isempty(filesin)
+        % Default to glob all
+        filesin = "**/*.m";
+    end
+    if isempty(outputfile) % Default to stdout
```
hsautessella, line 15: If `outputfile` is not specified then this errors. `if exists('outputfile', 'var') && isempty(outputfile)` would default to stdout as described

admin/lint_json.m
```diff
@@ -0,0 +1,52 @@
...
+        if fh == -1
+            error("MATLAB:FileOpenError", "Failed to open file %s", outputfile);
+        end
+        cleanup = onCleanup(@()(fclose(fh)));
+    end
+
+
+    files = [];
+    for i = 1:numel(filesin)
+        flist = dir(filesin{i});
```
hsautessella, line 28: Trying to run this file, I get a `Brace indexing is not supported for variables of this type.` error. On line 4 `filesin` is defined as a string, which is not `{}` indexable

admin/lint_json.m
```diff
@@ -0,0 +1,52 @@
+function lint_json(filesin, outputfile)
+    if isempty(filesin)
+        % Default to glob all
+        filesin = "**/*.m";
```
hsautessella, line 13: `filesep` gives a platform compatible file separator
```suggestion
        filesin = ['**', filesep, '*.m'];
```

CMakeLists.txt
hsautessella, line 28: Prefixing this file with `PACE_` would be more consistent with our other CMake function and provides a form of namespacing

hsautessella (changes requested):

admin/lint_json.m
oerc0122: `outputfile` can't be empty though, right? It's a required input arg?

hsautessella: I got the logic completely wrong here, sorry. `if ~exists('outputfile', 'var') || isempty(outputfile)` is more like what you'd want.

hsautessella: Also `exist` not `exists`... I should really try running this stuff before commenting 🙈

cmake/shared/CodeAnalysis.cmake
```diff
@@ -0,0 +1,24 @@
+add_custom_target(analyse
```
hsautessella, line 1: De we need to `EXCLUDE_FROM_ALL` these targets so they're not run on a plain `make`?

oerc0122: https://cmake.org/cmake/help/v3.15/command/add_custom_target.html

For custom targets you add to all, not exclude.

admin/lint_json.m
oerc0122: `fwrite` seems to be for writing raw data, we want formatted data. Unless I'm mistaken.
> `fwrite`
> Write data to binary file

https://uk.mathworks.com/help/matlab/ref/fwrite.html
https://uk.mathworks.com/help/matlab/ref/fprintf.html

cmake/shared/CodeAnalysis.cmake
hsautessella: Ah, you're right. I was running "Build Solution" on Visual Studio instead of building `ALL_BUILD`

admin/lint_json.m
hsautessella: I only said this because I was recently bitten by using `fprintf` to write a large text file, and it stopped writing after ~255,000 characters. But that may have been the way I had called it (I had `fprintf(fh, long_char_array)` not `fprintf(fh, "%s", long_char_array)`.

Happy to leave this as is

admin/lint_json.m
hsautessella: All arguments in a Matlab function are "optional". You'll only hit a "Not enough input arguments." error on the first use of an unspecified argument. If the intention is for both arguments to be required, but you specify empty strings for a default, that's OK, but needs to be in the function help.

cmake/shared/PACE_CodeAnalysis.cmake
```diff
@@ -5,7 +5,7 @@ add_custom_target(analyse
 add_custom_target(analyse-mlint
   COMMENT "- Performing MATLAB analysis (Mlint)..."
   BYPRODUCTS "${CMAKE_CURRENT_BINARY_DIR}/mlint.out"
-  COMMAND ${Matlab_MAIN_PROGRAM} -nodisplay -r "\"addpath('${Herbert_ROOT}/../admin');lint_json({'${CMAKE_SOURCE_DIR}/**/*.m'},'${CMAKE_CURRENT_BINARY_DIR}/mlint.json');exit\""
+  COMMAND ${Matlab_MAIN_PROGRAM} -nodisplay -r "\"addpath('${Herbert_ROOT}/admin');lint_json({'${CMAKE_SOURCE_DIR}/**/*.m'},'${CMAKE_CURRENT_BINARY_DIR}/mlint.json');exit\""
```
hsautessella, line 8: If you use `-batch` rather than `-r`, there's no need to call `exit`. It also means that we don't need to deal with cases where an error occurs in `mlint_json`. https://www.mathworks.com/help/matlab/ref/matlablinux.html#buknt99-1-option1optionN.

hsautessella: When I say "we don't need to deal with cases where an error occurs in `mlint_json`", I mean the Matlab process will be killed. As it is now, it's not clear to me how the Matlab process exits if `lint_json` errors.

admin/lint_json.m
```diff
@@ -0,0 +1,67 @@
...
+        if fh == -1
+            error("MATLAB:FileOpenError", "Failed to open file %s", outputfile);
+        end
+        cleanup = onCleanup(@()(fclose(fh)));
+    end
+
+    files = [];
+    for i = 1:numel(filesin)
+        flist = dir(filesin{i});
+        % Filter doc files
```
nickbattam-tessella, line 28: would it be helpful to pull this "filter" out as a helper function? It's well defined and simply takes an array and returns an array with (currently one) filter applied?

This function is starting to do a number of discrete processing steps which increases its complexity.

admin/lint_json.m
```diff
@@ -0,0 +1,67 @@
...
+        filter = arrayfun(@(x)(startsWith(x.name,'doc_')), flist);
+        filtered = flist(filter);
+        for j = 1:numel(filtered)
+            fprintf("Skipping: %s\n", filtered.name);
+        end
+        flist = flist(~filter);
+        files = [files; arrayfun(@(file)(fullfile(file.folder, file.name)), flist, 'UniformOutput', false)];
```
nickbattam-tessella, line 35: Is this an efficient way of building a potentially large list of files in a growing array? Other places in the code where this aggregation is performed we've built cell array of arrays and performed a single aggregation at the end

admin/lint_json.m
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    issues = struct('issues', {{}}, 'size', 0);
+    raw = checkcode(files, '-id');
+    for i = 1:numel(raw)
+        for j = 1:numel(raw{i})
+            raw{i}(j).fileName = files{i};
```
nickbattam-tessella, line 42: as you call `wng_compat` on the next line to parse this object, would it be cleaner to have all modifications done in there, passing `files{i}` as a second argument?

admin/lint_json.m
```diff
@@ -0,0 +1,67 @@
...
+            curr = wng_compat(raw{i}(j));
+            issues.issues = {issues.issues{:}, curr};
```
nickbattam-tessella, line 44: `issues.issues` is confusing naming

admin/lint_json.m
oerc0122: No, it's not an efficient way, but it's also not a performance critical part and is guaranteed to be a small array (unless we suddenly gain half a million `.m` files), so I'm not sure that's relevant. Simplicity of code takes priority here.

hsautessella (approved):
This looks good to me 👍

admin/lint_json.m
```diff
@@ -0,0 +1,73 @@
...
+        end
+        cleanup = onCleanup(@()(fclose(fh)));
+    end
+
+    files = [];
+    for i = 1:numel(filesin)
+        flist = dir(filesin{i});
+        % Filter doc files
+        flist = filter_list(flist, @(x)(startsWith(x.name,'doc_')));
+        flist = arrayfun(@(file)(fullfile(file.folder, file.name)), flist, 'UniformOutput', false)
```
hsautessella, line 30: Missing `;` at EOL

nickbattam-tessella (approved):

hsautessella (approved):

PR #293, opened 2021-02-10
Set JobStorageLocation on parpool MPI clusters

hsautessella:
Set Matlab Parpools' `JobStorageLocation` to `parallel_config.working_directory` in MPI framework.

This should stop the same directory being used by multiple different Matlab instances at the same time on Jenkins. 

`parallel_config.working_directory` evaluates to `JENKINS_WORKSPACE/build/test` on Jenkins, hence each executor should have its own (separate) copy of the folder. See [relevant Jenkins build](https://anvil.softeng-support.ac.uk/jenkins/job/PACE-neutrons/job/Herbert/job/Branch-Windows-10-2019b/28/console) where I print the value of `JobStorageLocation`.

Addresses https://github.com/pace-neutrons/Horace/issues/557


oerc0122 (approved):
If this is all that's needed, looks good.

PR #299, opened 2021-02-12
Serialise/Deserialise code modified to continue running

abuts:
when mex file is not available. 

Simple modification allows user to continue running the code if mex code for serialiser is not available. This makes the code consistent with all other places where the mex code is used. 

User will be nagged to do something about missing code (disable mex or compile mex)


hsautessella (commented):
Looks sensible to me, but do we need to cover the case where `force_mex` is on?

oerc0122 (approved):

hsautessella (approved):

PR #300, opened 2021-02-15
567 lint ignore herbert

oerc0122:
Add ability to exclude files from mlint parsing. Syntax
`mlint_json({'files', 'to', 'scan', '**/*.m'}, 'output', 'exclude', {'excluded', 'files', '**/doc_*'});`

### DIFF
--- a/admin/lint_json.m
+++ b/admin/lint_json.m
@@ -3,10 +3,14 @@ function lint_json(filesin, outputfile, varargin)
 %
 % Input:
 % ------
-% filesin              char array OR cell array of char arrays detailing files to parse
+% filesin              cell array of char arrays detailing files to parse
 %                          if filesin is empty will recurse from current working directory
 % outputfile           char array of filename to write output to (will overwrite)
 %                          if outputfile is empty will write to stdout
+%
+% Keyword arguments:
+% exclude              cell array of char arrays detailing files to exclude from parsing
+%
 
     p = inputParser;
     addOptional(p, 'filesin', {['**', filesep, '*.m']}, @iscellstr);
@@ -14,27 +18,31 @@ function lint_json(filesin, outputfile, varargin)
     addParameter(p, 'exclude', {}, @iscellstr);
     parse(p, filesin, outputfile, varargin{:});
 
-    if strcmp(p.Results.outputfile, '_screen')
+    filesin = p.Results.filesin;
+    outputfile = p.Results.outputfile;
+    exclude = p.Results.exclude;
+
+    if strcmp(outputfile, '_screen')
         fh = 1;
     else % Open file
-        fh = fopen(p.Results.outputfile,'w');
+        fh = fopen(outputfile,'w');
         if fh == -1
-            error("MATLAB:FileOpenError", "Failed to open file %s", p.Results.outputfile);
+            error("MATLAB:FileOpenError", "Failed to open file %s", outputfile);
         end
         cleanup = onCleanup(@()(fclose(fh)));
     end
 
     files = [];
-    for i = 1:numel(p.Results.filesin)
-        flist = dir(p.Results.filesin{i});
+    for i = 1:numel(filesin)
+        flist = dir(filesin{i});
         % Filter doc files
         flist = filter_list(flist, @(x)(startsWith(x.name,'doc_')));
         flist = arrayfun(@(file)(fullfile(file.folder, file.name)), flist, 'UniformOutput', false);
         files = unique([files; flist]);
     end
 
-    for i = 1:numel(p.Results.exclude)
-        excl = dir(p.Results.exclude{i});
+    for i = 1:numel(exclude)
+        excl = dir(exclude{i});
         excl = arrayfun(@(file)(fullfile(file.folder, file.name)), excl, 'UniformOutput', false);
         files = setdiff(files, excl);
     end
--- a/admin/lint_json.m
+++ b/admin/lint_json.m
@@ -9,7 +9,7 @@ function lint_json(filesin, outputfile, varargin)
 %                          if outputfile is empty will write to stdout
 
     p = inputParser;
-    addOptional(p, 'filesin', {['**', filesep, '*.m']});
+    addOptional(p, 'filesin', {['**', filesep, '*.m']}, @iscellstr);
     addOptional(p, 'outputfile', '_screen', @ischar);
     addParameter(p, 'exclude', {}, @iscellstr);
     parse(p, filesin, outputfile, varargin{:});
--- a/admin/lint_json.m
+++ b/admin/lint_json.m
@@ -1,4 +1,4 @@
-function lint_json(filesin, outputfile)
+function lint_json(filesin, outputfile, varargin)
 % Use mlint, and convert to a json for easy parsing by WNG
 %
 % Input:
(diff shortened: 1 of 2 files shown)


cmake/shared/PACE_CodeAnalysis.cmake
```diff
@@ -2,10 +2,14 @@ add_custom_target(analyse
...
 
+# Handle CMAKE list to matlab cell array
+string(JOIN "','" IGNORE_STRING ${PACE_MLINT_IGNORE})
+string(CONCAT IGNORE_STRING "{'" IGNORE_STRING "'}")
+
 add_custom_target(analyse-mlint
   COMMENT "- Performing MATLAB analysis (Mlint)..."
   BYPRODUCTS "${CMAKE_CURRENT_BINARY_DIR}/mlint.out"
-  COMMAND ${Matlab_MAIN_PROGRAM} -nodisplay -batch "\"addpath('${Herbert_ROOT}/admin');lint_json({'${CMAKE_SOURCE_DIR}/**/*.m'},'${CMAKE_CURRENT_BINARY_DIR}/mlint.json');exit\""
+  COMMAND ${Matlab_MAIN_PROGRAM} -nodisplay -batch "\"addpath('${Herbert_ROOT}/admin');lint_json({'${CMAKE_SOURCE_DIR}/**/*.m'},'${CMAKE_CURRENT_BINARY_DIR}/mlint.json','exclude',${IGNORE_STRING});exit\""
```
nickbattam-tessella, line 12: As the command string is now getting quite long ad complex would it be clear to build it as a variable outside the `add_custom_target` call?

For readability can this be split over multiple lines too?

admin/lint_json.m
```diff
@@ -8,27 +8,35 @@ function lint_json(filesin, outputfile)
...
+    addOptional(p, 'filesin', {['**', filesep, '*.m']}, @iscellstr);
+    addOptional(p, 'outputfile', '_screen', @ischar);
+    addParameter(p, 'exclude', {}, @iscellstr);
+    parse(p, filesin, outputfile, varargin{:});
+
+    if strcmp(p.Results.outputfile, '_screen')
         fh = 1;
     else % Open file
-        fh = fopen(outputfile,'w');
+        fh = fopen(p.Results.outputfile,'w');
```
nickbattam-tessella, line 20: Using data directly from `p.Results` smears the arg parser though the whole function. It would be cleaner to pull the variables we're interested in out of this structure.

admin/lint_json.m
```diff
@@ -8,27 +8,35 @@ function lint_json(filesin, outputfile)
...
     files = [];
-    for i = 1:numel(filesin)
-        flist = dir(filesin{i});
+    for i = 1:numel(p.Results.filesin)
+        flist = dir(p.Results.filesin{i});
         % Filter doc files
         flist = filter_list(flist, @(x)(startsWith(x.name,'doc_')));
         flist = arrayfun(@(file)(fullfile(file.folder, file.name)), flist, 'UniformOutput', false);
-        files = [files; flist];
+        files = unique([files; flist]);
```
nickbattam-tessella, line 33: is there an efficiency to doing this on each loop rather than once at the end?

cmake/shared/PACE_CodeAnalysis.cmake
```diff
@@ -2,10 +2,14 @@ add_custom_target(analyse
   COMMENT "Performing code analysis..."
   )
 
+# Handle CMAKE list to matlab cell array
+string(JOIN "','" IGNORE_STRING ${PACE_MLINT_IGNORE})
+string(CONCAT IGNORE_STRING "{'" IGNORE_STRING "'}")
```
nickbattam-tessella, line 7: Is it possible to perform this final wrap in `{'` ... `'}` as part of the command string you're building below?

admin/lint_json.m
```diff
@@ -8,27 +8,35 @@ function lint_json(filesin, outputfile)
...
+            error("MATLAB:FileOpenError", "Failed to open file %s", p.Results.outputfile);
         end
         cleanup = onCleanup(@()(fclose(fh)));
     end
 
     files = [];
-    for i = 1:numel(filesin)
-        flist = dir(filesin{i});
+    for i = 1:numel(p.Results.filesin)
+        flist = dir(p.Results.filesin{i});
```
nickbattam-tessella, line 29: you reassign the `flist` variable multiple times in this loop and change its data type (from filename to filepath). Can you use distinct variables?

cmake/shared/PACE_CodeAnalysis.cmake
```diff
@@ -2,10 +2,14 @@ add_custom_target(analyse
   COMMENT "Performing code analysis..."
   )
 
+# Handle CMAKE list to matlab cell array
+string(JOIN "','" IGNORE_STRING ${PACE_MLINT_IGNORE})
```
nickbattam-tessella, line 6: You're now dependent on this variable being set; that should be noted in the function header. There are good examples of this in `PACE_Version.cmake`

admin/lint_json.m
```diff
@@ -1,4 +1,4 @@
-function lint_json(filesin, outputfile)
+function lint_json(filesin, outputfile, varargin)
```
nickbattam-tessella, line 1: Valid `varargin` values need adding to the function header

cmake/shared/PACE_CodeAnalysis.cmake
oerc0122: You *can*, but it adds extra brackets making it less legible. It's down to choice, but I prefer
`'exclude',${IGNORE_STRING});exit`
to
`'exclude',{'${IGNORE_STRING}'};exit`

cmake/shared/PACE_CodeAnalysis.cmake
oerc0122: If unset, it will be a blank string, which is exactly what we want.

admin/lint_json.m
oerc0122: Depends, in principle if you have specified a lot of overlapping files, maybe. In general, it's unlikely. However, again, efficiency is really not the key here. Here, I think it clarifies that we are taking the elements from `flist` which are new to `files`  and discarding others.

admin/lint_json.m
```diff
@@ -28,7 +38,13 @@ function lint_json(filesin, outputfile)
         % Filter doc files
         flist = filter_list(flist, @(x)(startsWith(x.name,'doc_')));
         flist = arrayfun(@(file)(fullfile(file.folder, file.name)), flist, 'UniformOutput', false);
-        files = [files; flist];
+        files = unique([files; flist]);
+    end
+
+    for i = 1:numel(exclude)
+        excl = dir(exclude{i});
+        excl = arrayfun(@(file)(fullfile(file.folder, file.name)), excl, 'UniformOutput', false);
```
nickbattam-tessella, line 46: is it possible to restructure the `filelist - exludefiles` operation so you only need one call to the arrayfun on the filtered list to build build the full paths of the retained files? 

e.g. build an initial list of unique `file`s (retaining filename and folder information), using setdiff to remove each of the excluded folders/files, and then building an output list of the full paths?

admin/lint_json.m
```diff
@@ -1,18 +1,28 @@
...
 % Input:
 % ------
-% filesin              char array OR cell array of char arrays detailing files to parse
+% filesin              cell array of char arrays detailing files to parse
 %                          if filesin is empty will recurse from current working directory
 % outputfile           char array of filename to write output to (will overwrite)
 %                          if outputfile is empty will write to stdout
+%
+% Keyword arguments:
+% exclude              cell array of char arrays detailing files to exclude from parsing
```
nickbattam-tessella, line 12: What does "detailing files to" mean? Is that as regex, glob, filename, fullpaths, ...? 

(applied equally to the `filesin` parameter)

cmake/shared/PACE_CodeAnalysis.cmake
```diff
@@ -1,11 +1,60 @@
...
+
+Run mlint and cppcheck code analysis on Project
+
+Variables required by the module
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``${Herbert_ROOT}``
+This is provided by the `FindHerbert` module which must be loaded first
+
+``${PACE_MLINT_IGNORE}``
```
nickbattam-tessella, line 13: what is this parameter? it's not immediately obvious what form the content should take or where (if anywhere) it's defined.

cmake/shared/PACE_CodeAnalysis.cmake
```diff
@@ -16,7 +65,9 @@ if (cppcheck)
   add_custom_target(analyse-cppcheck
     COMMENT "- Performing C++ analysis (CppCheck)..."
     BYPRODUCTS "${CMAKE_CURRENT_BINARY_DIR}/cppcheck.xml"
-    COMMAND cppcheck --enable=all --inconclusive --xml --xml-version=2 -I "${CMAKE_SOURCE_DIR}/_LowLevelCode/cpp" "${CMAKE_SOURCE_DIR}/_LowLevelCode/" 2> "${CMAKE_CURRENT_BINARY_DIR}/cppcheck.xml"
+    COMMAND cppcheck --enable=all --inconclusive --xml --xml-version=2
+    -I "${CMAKE_SOURCE_DIR}/_LowLevelCode/cpp" "${CMAKE_SOURCE_DIR}/_LowLevelCode/"
```
nickbattam-tessella, line 69: can you indent the continuation lines for clarity?

PR #304, opened 2021-02-17
Add skip CI based on labels

oerc0122:
Add the ability to skip CI based on the `do-not-build` and `do-not-test` PR labels

### DIFF
--- a/tools/build_config/Jenkinsfile
+++ b/tools/build_config/Jenkinsfile
@@ -112,9 +112,9 @@ properties([
   ])
 ])
 
-def handle_test_status() {
+def handle_test_status(Boolean testing_done) {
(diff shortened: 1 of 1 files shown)


tools/build_config/Jenkinsfile
```diff
@@ -304,6 +316,11 @@ pipeline {
         ]
       )
 
+      archiveArtifacts(
```
nickbattam-tessella, line 319: can this be merged back into the other archive call, or is there clarity in keeping it separate?

oerc0122: I think there is clarity in separation as they're done on different stages.

cmarooney-stfc (approved):

PR #305, opened 2021-02-17
Only skip merge SHA on non-nightly, non-branch runs

oerc0122:


cmarooney-stfc (approved):

nickbattam-tessella:
Has this been tested for BRANCH and PR pipelines?

cmake/shared/PACE_Version.cmake
```diff
@@ -35,34 +35,39 @@ set(${PROJECT_NAME}_FULL_VERSION "${PROJECT_VERSION}")
...
 if(NOT "${_release_type}" STREQUAL "RELEASE")
-    find_package(Git QUIET)
-    execute_process(
-        COMMAND ${GIT_EXECUTABLE} rev-list --abbrev-commit --no-merges -n 1 HEAD
-        WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
-        RESULT_VARIABLE _res
-        OUTPUT_VARIABLE GIT_REVISION_SHA
-        OUTPUT_STRIP_TRAILING_WHITESPACE
+  find_package(Git QUIET)
+  if ("${_release_type}" STREQUAL "NIGHTLY" OR "${_release_type}" STREQUAL "BRANCH")
```
nickbattam-tessella, line 52: Would it be safer and clearer to make the skip use-case explicit as that is the exceptional rather than normal behaviour, i.e. reverse to logic
```
set(SKIP_MERGES = "") // or as an else
if (this-is-a-PR) 
  set(SKIP_MERGES = "--no-merge")
end
```

PR #307, opened 2021-02-18
Set defaults in Jenkinsfile to allow easy addition of new params

oerc0122:
- Set defaults via `get_param` function
- Move printing of `PR_LABELS` for general use and consistency

### DIFF
--- a/tools/build_config/Jenkinsfile
+++ b/tools/build_config/Jenkinsfile
@@ -87,7 +87,7 @@ properties([
       name: 'MATLAB_VERSION', trim: true
     ),
     string(
-      defaultValue: '3.7.2',
+      defaultValue: '3.19',
       description: 'The version of CMake to run the build with.',
       name: 'CMAKE_VERSION',
       trim: true
--- a/tools/build_config/Jenkinsfile
+++ b/tools/build_config/Jenkinsfile
@@ -295,7 +295,10 @@ pipeline {
       steps {
         script {
           if (isUnix()) {
-            sh './tools/build_config/build.sh --package'
+            sh '''
+              module load cmake/\$CMAKE_VERSION &&
+              ./tools/build_config/build.sh --package
+            '''
           }
           else {
             powershell './tools/build_config/build.ps1 -package'
--- a/tools/build_config/Jenkinsfile
+++ b/tools/build_config/Jenkinsfile
@@ -171,17 +171,7 @@ properties([
 ])
 
 
-if (BRANCH_NAME) {
-  currentBuild.description = "Branch: ${BRANCH_NAME}"
-} else if (env.PR_NUMBER) {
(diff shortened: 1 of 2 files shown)


tools/build_config/Jenkinsfile
```diff
@@ -273,10 +295,7 @@ pipeline {
       steps {
         script {
           if (isUnix()) {
-            sh '''
-              module load cmake/\$CMAKE_VERSION &&
-              ./tools/build_config/build.sh --package
-            '''
+            sh './tools/build_config/build.sh --package'
```
nickbattam-tessella, line 298: does this not require the CMAKE_VERSION set? package calls CPack which is part of CMake

tools/build_config/Jenkinsfile
```diff
@@ -182,10 +190,24 @@ pipeline {
     label env.AGENT
   }
 
+  environment {
+    MATLAB_VERSION = get_param('MATLAB_VERSION', get_matlab_release(env.JOB_BASE_NAME))
```
nickbattam-tessella, line 194: a comment here or in the get_param function would make it explicit that this will use the variable passed from Jenkins UI OR the default if that is empty.

nickbattam-tessella (approved):

tools/build_config/Jenkinsfile
```diff
@@ -60,57 +60,18 @@ def get_branch_name(String job_name) {
...
-    ),
-    string(
-      defaultValue: '1.77',
-      description: 'The version of CppCheck tooling to load to provide the code-style checks.',
-      name: 'CPPCHECK_VERSION',
-      trim: true
-    )
-  ])
-])
+String get_param(String param_name, String default_val) {
```
cmarooney-stfc, line 63: Checking that you have to move the parameters([ block after post-github_status. It means that the diff cannot compare parameters and show they are identical

oerc0122: I moved it because it puts all the functions in one place. One of the future tasks (distant, I suspect) is to move all the shared functions into a Jenkins libray. This just makes that one bit easier. The parameters are identical, but the defaults have been removed from all except the three which are required Jenkins params.

tools/build_config/Jenkinsfile
hsautessella: I'm also suspicious whether this will work without the `module load`

hsautessella (commented):
Looks good and a quick test seemed to give the correct behaviour r.e. parameter values.

Note that the builds of this PR, though they passed, skipped the test & package stages.

tools/build_config/Jenkinsfile
oerc0122: Didn't realise it'd gone. Whoops.

nickbattam-tessella (approved):

PR #311, opened 2021-02-24
Add configure Jenkins stage

oerc0122:
Update to keep in line with Horace's new configure stage as part of [Horace #588 addressing Horace #586](https://github.com/pace-neutrons/Horace/pull/588)


tools/build_config/Jenkinsfile
```diff
@@ -190,6 +190,28 @@ pipeline {
...
+            --cmake_flags \"-DHorace_RELEASE_TYPE=\$RELEASE_TYPE\" \
+            --matlab_release \$MATLAB_VERSION \
+            --print_versions --configure
+          '''
+      } else {
+        powershell '''
+           ./tools/build_config/build.ps1 -print_versions -configure
+                -cmake_flags \"-DHorace_RELEASE_TYPE=\$env:RELEASE_TYPE\" \
+                -matlab_release \$env:MATLAB_VERSION \
+                -vs_version \$env:VS_VERSION
```
hsautessella, line 211: Thanks for adding this flag. Can we add a ref to https://github.com/pace-neutrons/Horace/issues/593 in the PR description, so we know this is fixing the issue for Herbert?

tools/build_config/Jenkinsfile
```diff
@@ -190,6 +190,28 @@ pipeline {
...
+          module load gcc/\$GCC_VERSION &&
+          module load cppcheck/\$CPPCHECK_VERSION &&
+          ./tools/build_config/build.sh \
+            --cmake_flags \"-DHorace_RELEASE_TYPE=\$RELEASE_TYPE\" \
+            --matlab_release \$MATLAB_VERSION \
+            --print_versions --configure
+          '''
+      } else {
+        powershell '''
+           ./tools/build_config/build.ps1 -print_versions -configure
```
hsautessella, line 208: Could you split each flag onto a separate line? Just makes it easier to visually parse
```suggestion
           ./tools/build_config/build.ps1 \
             -print_versions \ 
             -configure \
```

(Similar for the equivalent bash commands)

hsautessella (commented):
Changes look sensible to me, but the [build logs](https://anvil.softeng-support.ac.uk/jenkins/job/PACE-neutrons/job/Herbert/job/PR-Scientific-Linux-7-2018b/666/console) suggest you have a missing `}` somewhere.

tools/build_config/Jenkinsfile
```diff
@@ -190,6 +190,32 @@ pipeline {
       }
     }
 
+    stage('Configure') {
+      if (isUnix()) {
+        sh '''
+          export PATH=${PATH}:~/.local/bin &&
```
nickbattam-tessella, line 196: is this PATH modification required for Herbert? This was added to support the Python/doc steps in  Horace and is not set in any other build step.

tools/build_config/build.ps1
```diff
@@ -25,6 +25,8 @@
   https://github.com/pace-neutrons/Herbert
 #>
 param (
+  # Run the Herbert build commands.
```
cmarooney-stfc, line 28: This comment looks like a cut-and-paste from the next one. Did you mean "configure commands"?

tools/build_config/Jenkinsfile
```diff
@@ -225,16 +250,12 @@ pipeline {
               module load cmake/\$CMAKE_VERSION &&
               module load matlab/\$MATLAB_VERSION &&
               module load gcc/\$GCC_VERSION &&
-              ./tools/build_config/build.sh --build \
-                  --cmake_flags "-DHerbert_RELEASE_TYPE=$RELEASE_TYPE" \
-                  --matlab_release \$MATLAB_VERSION
+              ./tools/build_config/build.sh --build
```
cmarooney-stfc, line 253: Be useful to know where the herbert release type and matlab version are now set.

oerc0122: Do you mean the CMake arg or the environment variable? The CMake is in the configure stage as one might hope with usual CMake setups. The environment var is unchanged.

nickbattam-tessella (approved):
Looks to resolve the passing of variables and shell configuration issues.

The MATLAB version used in (Windows) build pipeline doesn't match the passed flag.
The VS version looks to match, it's unclear if that is happenstance or a consequence of the flag being parsed and applied correctly.

oerc0122:
> Looks to resolve the passing of variables and shell configuration issues.
> 
> The MATLAB version used in (Windows) build pipeline doesn't match the passed flag.
> The VS version looks to match, it's unclear if that is happenstance or a consequence of the flag being parsed and applied correctly.

The MATLAB version *printed* doesn't match. The version used does. I will take a look.

tools/build_config/Jenkinsfile
```diff
@@ -213,6 +213,35 @@ pipeline {
       }
     }
 
+    stage('Configure') {
+      steps {
+        script {
+          if (isUnix()) {
+            sh '''
+          module load cmake/\$CMAKE_VERSION &&
+          module load matlab/\$MATLAB_VERSION &&
```
hsautessella, line 222: You've lost some indentation here

tools/build_config/Jenkinsfile
```diff
@@ -213,6 +213,35 @@ pipeline {
...
+          ./tools/build_config/build.sh \
+            --cmake_flags \"-DHorace_RELEASE_TYPE=\$RELEASE_TYPE\" \
+            --matlab_release \$MATLAB_VERSION \
+            --print_versions \
+            --configure
+          '''
+          } else {
+            powershell '''
+           ./tools/build_config/build.ps1 \
+                -cmake_flags \"-DHorace_RELEASE_TYPE=\$env:RELEASE_TYPE\" \
```
hsautessella, line 234: We should have an indent here too

tools/build_config/build.ps1
```diff
@@ -139,6 +130,7 @@ function New-CMake-Generator-Command {
 
 function Write-Versions {
   Write-Output "$(cmake --version)"
+  Write-Output "Matlab: $($(Get-Command matlab.exe).Source)"
```
hsautessella, line 133: Following our discussions on Slack, should we be printing this? The Matlab on the path may not be the Matlab that CMake picks up (since it searches the registry). This is a difference to Linux, where CMake will pick whichever Matlab is on the path

hsautessella (approved):
Couple very nitpicky comments on some lost indentation and a question on printing Matlab version. Otherwise, looks good

tools/build_config/build.ps1
nickbattam-tessella: why isn't this the correct path? how does CTest run with the correct MATLAB if the console can't find it? How is that path defined??

hsautessella: By default, at configure time, CMake will pick the latest version of Matlab out of the registry. All the data needed to run CTest is generated in the CMake configure stage, including the path to the Matlab it found. Have a look inside `build/_test/CTestTestfile.cmake`, it contains the full test commands, including the absolute path to the Matlab it found at configure time.

We can specify what version of Matlab CMake looks for by calling `find_package(Matlab EXACT ${_version})` - this is done in `PACE_FindMatlab` in Herbert.

oerc0122: I'm not sure we need any of them besides the cmake version as they're reported in the CMake configure

hsautessella: I think I agree. We should at least remove the Matlab one as it is potentially misleading

hsautessella (approved):

cmarooney-stfc (approved):

PR #314, opened 2021-02-26
Clarify which pipelines vars affect

oerc0122:


tools/build_config/Jenkinsfile
```diff
@@ -133,19 +133,19 @@ properties([
   parameters([
     string(
       defaultValue: get_branch_name(env.JOB_BASE_NAME),
-      description: 'The name of the branch to build.',
+      description: 'The name of the branch to build. *',
```
hsautessella, line 136: This parameter is not mandatory on PR builds (they get the branch name from the GitHub webhook JSON)

oerc0122: That's true, but given that's the exception and not the rule. I would also say that as that's handled automatically 99% of the time it's probably safer to make people aware that in general it should be specified.

hsautessella: On rebuilding a PR though, the developer will be shown the parameter entry page: the `BRANCH_NAME` parameter will be empty and shown as mandatory. This could be misleading, particularly as I think the parameter would just be totally ignored in a PR build.

As this is, we're assuming the developer has prior knowledge that the parameter is not required for PR builds. But adding something like `* (parameter ignored in PR builds)` would clear up any confusion.

oerc0122: Good point, would you prefer something like that to keep in line?

hsautessella (approved):
Nice quality of life change :+1:

nickbattam-tessella (approved):

PR #318, opened 2021-03-04
range_add_border function and unit tests for it

abuts:
This is small function, moved from Horace, modified to remove duplicated code and unit tested to ensure its correctness.
May be of interest to the people, who reviewed Horace  [#591](https://github.com/pace-neutrons/Horace/issues/591) and complained about this function

### DIFF
--- a/_test/test_utilities/test_range_add_border.m
+++ b/_test/test_utilities/test_range_add_border.m
@@ -1,0 +1,91 @@
+classdef test_range_add_border < TestCase
+    % Unit tests to check range_add_border function
+    
+    properties
+    end
+    
+    methods
+        function ps = test_range_add_border(varargin)
+            if nargin>0
+                name = varargin{1};
+            else
+                name = 'test_range_add_border ';
+            end
+            ps = ps@TestCase(name);
+        end
+        %
+        function test_zero_border_do_nothing(~)
+            range = [-1*ones(1,4);ones(1,4)];
+            range_out = range_add_border(range,0);
+            assertEqual(range,range_out);
+        end
+        %
+        function test_no_tol_eq_eps_tol(~)
+            range = [-1*ones(1,4);ones(1,4)];
+            range_no  = range_add_border(range);
+            range_eps = range_add_border(range,-eps);
+            assertEqual(range_no,range_eps );
+        end
+        %
+        function test_zero_width(~)
+            range = [-1,1,-1,1;-1,1,-1,1];
+            range_zero  = range_add_border(range);
+            fe = eps;
+            ref_range = [-1-fe,1-fe,-1-fe,1-fe;
+                -1+fe,1+fe,-1+fe,1+fe];
+            assertEqual(range_zero,ref_range);
+        end
+        %
+        function test_near_zero(~)
+            range = [0,0,0,0;0,1,0,1];
+            range_zero  = range_add_border(range);
+            fe = eps;
+            ref_range = [-fe,-fe,-fe,-fe;
+                fe,1+fe,fe,1+fe];
+            assertEqual(range_zero,ref_range);
+        end
+        %
+        function test_large_value_relerr(~)
+            range = [-100,-100;100,100];
+            range_calc  = range_add_border(range);
+            fe = eps;
+            ref_range = [-100-100*fe,-100-100*fe;...
+                100+100*fe,100+100*fe];
+            assertEqual(range_calc,ref_range);
+        end
+        %
+        function test_large_value_abserr_ignored(~)
+            range = [-100,-100;100,100];
+            fe = eps;
+            range_calc  = range_add_border(range,fe);
+            assertEqual(range_calc,range);
+        end
+        function test_eps_border_is_valid(~)
+            range = [-1,1;-1,1];
+            range_calc  = range_add_border(range);
+            range_rat = range_calc(1,:)./range_calc(2,:);
+            range_ref = range(1,:)./range(2,:);
+            difr = abs(range_rat-range_ref)-2*eps;
+            zer = zeros(1,2);
+            assertEqual(difr ,zer);
+        end
+        
+        function test_eps_abserr_on_large_border_is_relerr(~)
+            range = [-100,100,-1,-10;-100,100,1,10];
+            range_calc  = range_add_border(range,eps);
+            
+            assertEqual(range_calc(1,1),-100-100*eps);
+            assertEqual(range_calc(2,1),-100+100*eps);
+            assertEqual(range_calc(1,2),100-100*eps);
+            assertEqual(range_calc(2,2),100+100*eps);
+            assertEqual(range_calc(1,3),-1-eps);
+            assertEqual(range_calc(2,3),1+eps);
+            assertEqual(range_calc(1,4),-10);
+            assertEqual(range_calc(2,4),10);
+            
+        end
+        
+        
+    end
+end
+
--- a/_test/test_utilities/test_range_add_border.m
+++ b/_test/test_utilities/test_range_add_border.m
@@ -20,7 +20,7 @@ classdef test_range_add_border < TestCase
             assertEqual(range,range_out);
         end
         %
-        function test_no_tol_eq_eps_tol(~)
+        function test_no_tol_eq_rel_eps_tol(~)
             range = [-1*ones(1,4);ones(1,4)];
             range_no  = range_add_border(range);
             range_eps = range_add_border(range,-eps);
@@ -93,7 +93,16 @@ classdef test_range_add_border < TestCase
             assertEqual(range_calc(2,3),1+eps);
             assertEqual(range_calc(1,4),-10);
             assertEqual(range_calc(2,4),10);
-            
+        end
+        function test_invalid_argumemt_nan(~)
+            range = [-100,100,-1,-10;-100,100,1,10];
+            assertExceptionThrown(@()range_add_border(range,nan),...
+                'RANGE_ADD_BORDER:invalid_argument');
+        end
+        function test_invalid_argumemt_symbol(~)
+            range = [-100,100,-1,-10;-100,100,1,10];
+            assertExceptionThrown(@()range_add_border(range,'s'),...
+                'RANGE_ADD_BORDER:invalid_argument');
         end
     end
 end
--- a/_test/test_utilities/test_range_add_border.m
+++ b/_test/test_utilities/test_range_add_border.m
@@ -63,14 +63,23 @@ classdef test_range_add_border < TestCase
         function test_eps_border_is_valid(~)
             range = [-1,1;-1,1];
             range_calc  = range_add_border(range);
+            % -1 ranges ratio is expected to be 1+2eps and 
+            % +1 ranges ration is expected to be 1-2*eps
             range_rat = range_calc(1,:)./range_calc(2,:);
             range_ref = range(1,:)./range(2,:);
-            difr = abs(range_rat-range_ref)-2*eps;
+            % is this estimate indeed correct?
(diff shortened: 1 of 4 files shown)


_test/test_utilities/test_range_add_border.m
```diff
@@ -0,0 +1,91 @@
...
+            ps = ps@TestCase(name);
+        end
+        %
+        function test_zero_border_do_nothing(~)
+            range = [-1*ones(1,4);ones(1,4)];
+            range_out = range_add_border(range,0);
+            assertEqual(range,range_out);
+        end
+        %
+        function test_no_tol_eq_eps_tol(~)
```
nickbattam-tessella, line 23: `...eq_rel_eps_tol` ? It's important that it's a relative eps not abs eps as they're different

_test/test_utilities/test_range_add_border.m
```diff
@@ -0,0 +1,91 @@
...
+            range = [-1,1;-1,1];
+            range_calc  = range_add_border(range);
+            range_rat = range_calc(1,:)./range_calc(2,:);
+            range_ref = range(1,:)./range(2,:);
+            difr = abs(range_rat-range_ref)-2*eps;
+            zer = zeros(1,2);
+            assertEqual(difr ,zer);
+        end
+        
+        function test_eps_abserr_on_large_border_is_relerr(~)
```
nickbattam-tessella, line 73: I don't understand what this test is testing

_test/test_utilities/test_range_add_border.m
```diff
@@ -0,0 +1,91 @@
+classdef test_range_add_border < TestCase
```
nickbattam-tessella, line 1: why define `fe = eps` in most of these tests?

_test/test_utilities/test_range_add_border.m
```diff
@@ -0,0 +1,91 @@
...
+        function test_large_value_relerr(~)
+            range = [-100,-100;100,100];
+            range_calc  = range_add_border(range);
+            fe = eps;
+            ref_range = [-100-100*fe,-100-100*fe;...
+                100+100*fe,100+100*fe];
+            assertEqual(range_calc,ref_range);
+        end
+        %
+        function test_large_value_abserr_ignored(~)
```
nickbattam-tessella, line 57: why is an eps sized abs border ignored?

_test/test_utilities/test_range_add_border.m
```diff
@@ -0,0 +1,91 @@
...
+            assertEqual(range_calc,range);
+        end
+        function test_eps_border_is_valid(~)
+            range = [-1,1;-1,1];
+            range_calc  = range_add_border(range);
+            range_rat = range_calc(1,:)./range_calc(2,:);
+            range_ref = range(1,:)./range(2,:);
+            difr = abs(range_rat-range_ref)-2*eps;
+            zer = zeros(1,2);
+            assertEqual(difr ,zer);
```
nickbattam-tessella, line 70: what has this tested?

_test/test_utilities/test_range_add_border.m
nickbattam-tessella: what is large? when does the behaviour from adding eps to not adding eps happen?

herbert_core/utilities/general/range_add_border.m
```diff
@@ -0,0 +1,60 @@
...
+%
+%
+if nargin == 1 % add epsilon-sized border
+    tol = -eps;
+end
+%
+ndim=size(range_in,2);
+if tol==0
+    range=range_in;
+    return
```
nickbattam-tessella, line 33: with a return here the next line can be `if` not `elseif`, otherwise there's no point in the return since there is no code outside the elseif/else block.

herbert_core/utilities/general/range_add_border.m
```diff
@@ -0,0 +1,60 @@
...
+    range=range_in+tol*([-ones(1,ndim);ones(1,ndim)]);
+    zero_width = abs(range_in(2,:)-range_in(1,:))<tol;
+    if any(zero_width)
+        sig_range = sign(range_in);
+        min_border = 1-tol*sig_range(1,:);
+        max_border = 1+tol*sig_range(2,:);
+        range(1,zero_width)=range(1,zero_width).*min_border(zero_width);
+        range(2,zero_width)=range(2,zero_width).*max_border(zero_width);
+    end
+elseif tol<0
```
nickbattam-tessella, line 44: this is `else` as you've had `==0`, `>0` cases already

herbert_core/utilities/general/range_add_border.m
```diff
@@ -0,0 +1,60 @@
...
+%   if tol is omitted, the tol assumed to be equal -4*eps where eps
+%   is minimal value such as 1+eps~=1;
+%
+%   tol         Control size of border:
+%               tol=0   No border
+%               tol>0   Absolute value of thickness of border
+%               tol<0   Relative size as a proportion of the range along
+%                       each axis. If the range is zero, absolute tol value
+%                       is used.
+% if tol is omitted the routine adds 4*epsilon sized border
```
nickbattam-tessella: this is nolonger the case, you're adding `eps` not `4 * eps`

_test/test_utilities/test_range_add_border.m
abuts: because it was 4eps initially. May be reverted back if eps is insufficient

_test/test_utilities/test_range_add_border.m
abuts: it is known issue with eps and large numbers as eps is minimal value so 1+eps ~=1; 10+eps == 10. This is expected behavior -- tested alongside with large absolute value, zero size range, as for  zero size range it still has to be eps between the ranges.

abuts: it just tests expected behavior. Large values here will probably start from 2, but may be processor-dependent

_test/test_utilities/test_range_add_border.m
abuts: as test name states: test_eps_border_is_valid, i.e. if I add small borders to min and max values in the case of small range, they are added correctly and treated correctly.  I will add couple of words to the test description

_test/test_utilities/test_range_add_border.m
abuts: I've added description -- it tests separate branch of code

herbert_core/utilities/general/range_add_border.m
```diff
@@ -0,0 +1,58 @@
...
+    range=range_in;
+    return
+elseif tol>0
+    range=range_in+tol*([-ones(1,ndim);ones(1,ndim)]);
+    zero_width = abs(range_in(2,:)-range_in(1,:))<tol;
+    if any(zero_width) % abserr for zero-width border is defined as relerr
+        sig_range = sign(range_in);
+        min_border = 1-tol*sig_range(1,:);
+        max_border = 1+tol*sig_range(2,:);
+        range(1,zero_width)=range(1,zero_width).*min_border(zero_width);
```
cmarooney-stfc, line 41: This is very clever code and I had to walk through it in the debugger to find out it was working. A description of how it works would save a future developer some time when debugging.

cmarooney-stfc (changes requested):

_test/test_utilities/test_range_add_border.m
```diff
@@ -0,0 +1,91 @@
...
+                fe,1+fe,fe,1+fe];
+            assertEqual(range_zero,ref_range);
+        end
+        %
+        function test_large_value_relerr(~)
+            range = [-100,-100;100,100];
+            range_calc  = range_add_border(range);
+            fe = eps;
+            ref_range = [-100-100*fe,-100-100*fe;...
+                100+100*fe,100+100*fe];
```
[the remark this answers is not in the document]
abuts: I have intentionally wrote this way, to be different from the code inside the function (where it is multiplied by the value)

_test/test_utilities/test_range_add_border.m
```diff
@@ -0,0 +1,91 @@
...
+            range_calc  = range_add_border(range,eps);
+            
+            assertEqual(range_calc(1,1),-100-100*eps);
+            assertEqual(range_calc(2,1),-100+100*eps);
+            assertEqual(range_calc(1,2),100-100*eps);
+            assertEqual(range_calc(2,2),100+100*eps);
+            assertEqual(range_calc(1,3),-1-eps);
+            assertEqual(range_calc(2,3),1+eps);
+            assertEqual(range_calc(1,4),-10);
+            assertEqual(range_calc(2,4),10);
```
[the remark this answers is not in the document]
abuts: because abserr on large value is ignored (expected behavior)

herbert_core/utilities/general/range_add_border.m
```diff
@@ -0,0 +1,58 @@
...
+% Add a small border to a range.
+%
+%   >> range=range_add_border(range_in)
+%   >> range=range_add_border(range_in,tol)
+%
+% Input:
+% ------
+%   range_in   Range of data (2xn array)
+%               [u1_min,u2_min,...;u1_max,u2_max,...]
+%   if tol is omitted, the tol assumed to be equal -eps where eps
```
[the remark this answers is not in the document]
abuts: let's not be ridiculous.  NaN-s remain Nan-s and let they stay this way. Routine should deal with the situations where one may occur, not with any trouble in the word. 

To is ok.

herbert_core/utilities/general/range_add_border.m
abuts: difficult to explain what the code explain the best but I have tried. At least I've expressed intentions.

herbert_core/utilities/general/range_add_border.m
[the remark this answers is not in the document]
cmarooney-stfc: @abuts At this point I'm correcting the English. You can either have "the tol is assumed to equal -eps" or "the tol is assumed to be equal to -eps". I capitalise words when I want to emphasize that they should be added or removed.

For the elseif comment above at line 46, I think it is good that if/elseif/else blocks should have an else so that the logic doesn't accidentally skip without giving notice. NaN was just a possible example. You're welcome either to follow Nick and have the tol<0 branch as a simple else or to keep your tol<o branch and add a full else afterwards with an error. I was just trying to split the difference between the two of you and let you go with his completeness check and your preferred if structure.

herbert_core/utilities/general/range_add_border.m
abuts: not sure it is very useful complication but can be added

herbert_core/utilities/general/range_add_border.m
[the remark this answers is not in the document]
abuts: I've hopefully modified the code within these ideas and have mentioned the changes above, though still not sure these checks add much value. Its developer function, used rarely and mainly internally and thorough check of all possible arguments in the world does not look reasonable though makes no harm (within current usage)

_test/test_utilities/test_range_add_border.m
```diff
@@ -0,0 +1,109 @@
...
+            assertExceptionThrown(@()range_add_border(range,nan),...
+                'RANGE_ADD_BORDER:invalid_argument');
+        end
+        function test_invalid_argumemt_symbol(~)
+            range = [-100,100,-1,-10;-100,100,1,10];
+            assertExceptionThrown(@()range_add_border(range,'s'),...
+                'RANGE_ADD_BORDER:invalid_argument');
+        end
+    end
+end
```
oerc0122, line 108: What happens if range is empty? i.e. `range = []`

oerc0122 (approved):

PR #324, opened 2021-03-11
Disable unstable tests

oerc0122:
Disabled:
test_mpi/test_job_dispatcher_mpiexec

Until such a time as it can be stabilised


nickbattam-tessella (approved):
Disabled test should be added to `_test/disabled_tests.md`

oerc0122:
Are we still maintaining disabled_tests.md? It was not correct when I last did checks while implementing skipTest. I thought part of the point of skipTest is that we now had logging on which tests were disabled and why. I guess it could still be helpful.

hsautessella (changes requested):
This isn't the test that failing I don't think. 

The tests this class runs are actually defined in the parent class - I believe those are the ones that are failing. You'll need to override those and skip them

hsautessella:
Also, these tests consistently pass on Linux, is it desirable to skip them? Would just skipping on Windows be better, or more confusing?

nickbattam-tessella:
Disabling the tests on all platforms will be simpler. 

#329 has been created as a high-priority ticket to resolve the underlying problems and enable the tests to be run again on all platforms.

oerc0122 (commented):
Looks good provided it passes, and I appear to be getting all the credit. 

I can't approve my own PRs, so this'll have to do.

nickbattam-tessella (approved):
Are the boilerplate overrides in `_mpiexec`, `_parool` and `_herbert` for the four test functions required?

hsautessella:
> _Are the boilerplate overrides in _mpiexec, _parool and _herbert for the four test functions required?_

The overrides are just boilerplate as far as I can tell; all they're doing is calling the parent class's method.

PR #325, opened 2021-03-12
Change Jenkins config path

oerc0122:
Change the path to the config to be unique for each Jenkins instance.


herbert_core/configuration/@config_store/config_store.m
```diff
@@ -31,13 +31,13 @@
             if is_virtual
                 newStore.config_folder_name_ = ['mprogs_config_',type];
             end
-            [is_virtual,build_name] = is_jenkins();
+            [is_virtual,build_name,workspace] = is_jenkins();
```
cmarooney-stfc, line 34: Looking at the use of is_virtual, it appears to really mean - is running on an instance of the platform (`is_platform`) for the call that returned it. For clarity it may be easier to rename these as `is_on_idaaas`, or `is_on_jenkins`, etc

oerc0122: I think this is purely to maintain syntactic compatibility with other `is*` functions: `isstruct`, etc. I don't think there's really any ambiguity here warranting a name change throughout.

herbert_core/admin/is_jenkins.m
```diff
@@ -1,22 +1,25 @@
...
+job_name = '';
+workspace = '';
 for i = 1:length(jenkins_env_vars)
     if isempty(getenv(jenkins_env_vars{i}))
         is_jenkins_pc = false;
         return;
     end
 end
 is_jenkins_pc = true;
+workspace = getenv('WORKSPACE');
```
nickbattam-tessella, line 19: for consistency with the JOB_NAME test below should this be wrapped if an `if nargout > 2` ?

herbert_core/admin/is_jenkins.m
```diff
@@ -1,22 +1,25 @@
-function [is_jenkins_pc,job_name] = is_jenkins()
+function [is_jenkins_pc,job_name,workspace] = is_jenkins()
 % Attempt to determine if the PC we're running on is a Jenkins server.
 %   This is detected through environment variables that are always set by
 %   Jenkins.
 %
 % Returns:
 % is_jenkins -- true if the progam is running on Jenkins
-% job_name   -- the value of the Jenkins variable "JOB_NAME", 
+% job_name   -- the value of the Jenkins variable "JOB_NAME",
```
nickbattam-tessella, line 8: missing `workspace` return value from the header comments

herbert_core/configuration/@config_store/config_store.m
nickbattam-tessella: is this workspace/config fix ONLY required when running as part of a Jenkins job, or is this more general behaviour change?

abuts (approved):
As I have fixed the tests, I am approving it :). Should work. Will see if it fixes all issues.

oerc0122 (commented):
Looks good to me.

PR #327, opened 2021-03-15
Add struct_to_named_args_cell function

hsautessella:
This function wraps/imitates the Matlab funcion namedargs2cell which was
introduced in release R2019b.

This function is useful for converting unmatched parser arguments
(stored in a struct) to a cell array so that unmatched arguments can be
passed down to a subsequent function call.

Function required by https://github.com/pace-neutrons/horace/issues/550


oerc0122 (approved):
Does what it says it's going to. Do we also want an inverse function? Or is that just:
cellArr = struct_to_named_args_cell(input)
retrieve = struct(cellArr{:})

cmarooney-stfc (approved):
Looks fine

Issue #330, opened 2021-03-16
Horace build flag passed to Herbert in Jenkins


hsautessella:
A Horace CMake flag is being passed to Herbert builds in the `Configure` Jenkins stage:

```txt
Running CMake configure step...
+ cmake "...\workspace\PACE-neutrons\Herbert\PR-Windows-10-2018b" -G "Visual Studio 15 2017 Win64" -DBUILD_TESTS=ON -DMatlab_RELEASE=R2018b -DHorace_RELEASE_TYPE=pull_request
```
This should read
```txt
Running CMake configure step...
+ cmake "...\workspace\PACE-neutrons\Herbert\PR-Windows-10-2018b" -G "Visual Studio 15 2017 Win64" -DBUILD_TESTS=ON -DMatlab_RELEASE=R2018b -DHerbert_RELEASE_TYPE=pull_request
```

This just requires fixing the command in the Jenkinsfile.

PR #331, opened 2021-03-16
Horace build flag passed to Herbert in Jenkins

hsautessella:
Fix the CMake flag in the Jenkinsfile. This flag is important when setting package names. Whilst it was incorrect, nightly packages did not have the date appended to the name.

Addresses #330


oerc0122 (approved):
Sorry, must've been a copy-paste error from the changes.

PR #334, opened 2021-03-19
Return MException from assertExceptionThrown

hsautessella:
Having access to the exception that was thrown can be useful for checking error messages/causes.


oerc0122 (approved):
Looks fine to me.

nickbattam-tessella (commented):
Change looks fine. As a change to a core library function example tests demonstrating that it works as expected for the three basic use cases: no exception, wrong exception, correct exception would be helpful.

nickbattam-tessella (approved):

PR #338, opened 2021-03-25
bug in box intersect

abuts:
this is small fix to the issue, identified during the work with symmetrisation. The unit tests which identify the issue are written and fixture to the intersect deals with these tests.


nickbattam-tessella (commented):
Rather than normalising the length of the `normal` vector in each `intersectnD` function, can that be done once in the `intersection` method, in the same way `edg` is normalised at the start of that function?

oerc0122 (commented):
Bounding boxes already exist in MATLAB, do we want to reinvent the wheel?

https://uk.mathworks.com/help/matlab/2-d-and-3-d-bounding-regions.html

nickbattam-tessella:
> Bounding boxes already exist in MATLAB, do we want to reinvent the wheel?
> 
> https://uk.mathworks.com/help/matlab/2-d-and-3-d-bounding-regions.html

Does this extend to 4D?

oerc0122:
> > Bounding boxes already exist in MATLAB, do we want to reinvent the wheel?
> > https://uk.mathworks.com/help/matlab/2-d-and-3-d-bounding-regions.html
> 
> Does this extend to 4D?

Hm. Apparently not, that can be ignored then.

abuts:
@nickbattam-tessella: Not sure about normal -- now I am normalizing it as soon as it has been build and I am building normals in different way for different geometry.  Its probably possible to build normal using some generic algorithm, but I can not easily find one. Minimal efforts to normalization would be to put it immediately after nomal construction -- as it is done now. Can not find better place to put it.

abuts:
> 
> 
> > > Bounding boxes already exist in MATLAB, do we want to reinvent the wheel?
> > > https://uk.mathworks.com/help/matlab/2-d-and-3-d-bounding-regions.html
> > 
> > 
> > Does this extend to 4D?
> 
> Hm. Apparently not, that can be ignored then.

in addition to that, what exist is not exactly what I am doing here. We have a boxes, defined by min/max points and I am looking at intersections between these boxes. What is in Matlab is not exactly this and I do not know how to easily use existing Matlab algorithms for our purposes. I had a look on what Matlab has before commissioning this work -- have not found anything useful.

herbert_core/graphics/@IX_dataset_3d/sliceomatic.m
```diff
@@ -81,6 +81,9 @@
 
 [xlabel,ylabel,zlabel]=make_label(w);
 clim = [min(w.signal(:)) max(w.signal(:))];
+if clim(2) == clim(1)
```
nickbattam-tessella, line 84: I can't see any tests for this failure mode.

abuts: it fails in GUI and I do not know how to test it.

nickbattam-tessella: As this is limits for plotting, would it be better to set the limits as `clim(1) - 0.5` and `clim(1) + 0.5` in the case `clim(1) == clim(2)`? That would place the single data value in the centre of the range rather than at one edge

(same for the sliceomaticfigure) fix

abuts: yes, probably. I will make it consistent whith 2D plottings where +-1  is used in the similar situation.

PR #339, opened 2021-03-26
Update herbert on template

hsautessella:
Use a CMake configurable template instead of a dummy path in the `herbert_on` template. This makes the find/replace more straight-forward in CMake. It now requires only one `configure_file` call, rather than reading the file, performing a regex replace, and writing the file back out.

This commit also preps incoming Horace changes for pace-neutrons/Horace#639.


admin/CMakeLists.txt
```diff
@@ -1,21 +1,15 @@
...
+# parallel workers can access herbert_on/worker scripts.
+# Note that LOCAL_INIT_DIR is defined in the top-level CMakeLists
+configure_file("herbert_on.m.template" "${LOCAL_INIT_DIR}/herbert_on.m")
 if(${BUILD_TESTS})
-    configure_file("worker_4tests.m.template" "${WORKER_CONFIG_DIR}/worker_4tests.m")
+    configure_file("worker_4tests.m.template" "${LOCAL_INIT_DIR}/worker_4tests.m")
 endif()
-configure_file("worker_v2.m.template" "${WORKER_CONFIG_DIR}/worker_v2.m")
-
-set(_path_regex "[a-zA-Z0-9/\\\\_\\-\\.: ]+")
```
oerc0122, line 9: For the record in RegEx within ranges (`[]`) you don't need to escape (except `\` and `-` if it is non-terminal), the following RegEx should work just as well and be more readable:
`"[a-zA-Z0-9/_.: \\-]+"`
(N.b. the `-` is moved to the end to not specify a range)

oerc0122 (approved):
Look fine, if it runs

nickbattam-tessella (approved):

hsautessella:
`do-not-merge` tag is on for now, as this could break Horace builds if merged before https://github.com/pace-neutrons/Horace/pull/640

cmarooney-stfc (commented):
No particular things I wanted to comment on here. Just a sense that old usages of parameters are being replaced by other parameters, but these changes don't include definitions of these parameters. Checking that they are existing parameters that are just being repurposed here.

PR #340, opened 2021-04-06
Add a lazy fix to handle MATLAB strings in equal_to_tol

oerc0122:


abuts (approved):
equal_to_toll uses 'ignore_strings' option as `sqw` file stores path inside and path are usually different on different machines. Not sure what issue this commit solves otherwise, but do not see a harm in doing what is done here

oerc0122:
Basically it's the different between a charstring (`'a'`) and a MATLAB string (`"a"`), but either way it shouldn't error out.

PR #341, opened 2021-04-07
Add isindex to handle array index validation

oerc0122:


herbert_core/utilities/general/isindex.m
```diff
@@ -0,0 +1,3 @@
+function ok = isindex(vec)
+    ok = isvector(vec) && (islogical(vec) || (all(vec > 0 & all(floor(vec) == vec))));
+end
```
cmarooney-stfc, line 3: 1. Surprised not to see isindex used. Makes it unclear whether it is necessary and what use case it would fill.
2. BY inspection, ok==T => vec is a 1D array, possibly empty, and is either logical or a collection of positive integers. It is not clear what makes this an index for you.
3. matlab help says the additional test isnumeric wards off char input, and replacing floor(x)==x by mod(x,1)==0 wards off Inf and Nan.
4. Presuming your cleanup of islognumscalar has nothing to do with isindex
4.

oerc0122: So this is a port of Harry's `is_positive_int_vector_or_logical_vector` which is being used to verify indices passed through to the `PixelData` `get_x` functions are valid indices for what he's doing. 

My choice of name is to reflect the usage rather than the exact check. 

I will update the function to reflect points in 3.

The cleanup is because at first I thought this was a vector variant of `islognumscalar`

PR #350, opened 2021-04-20
Documentation and minor code improvements to parsing and checking

tgperring:
Help edited to clarify the use of the various lookup options. In a few cases lines of code added to parse input argumnets more robustly. No changes to functionality.

### DIFF
--- a/herbert_core/utilities/classes/@object_lookup/object_array.m
+++ b/herbert_core/utilities/classes/@object_lookup/object_array.m
@@ -1,7 +1,8 @@
-function A = object_array (obj, iarray)
+function A = object_array (obj, varargin)
 % Return a given object array from the original set of arrays
 %
-%   >> obj = object_array (iarray)
+%   >> obj = object_array (obj, iarray)
+%   >> obj = object_array (obj)             % OK if only one original array
 %
 % Input:
 % ------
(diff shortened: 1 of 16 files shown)


herbert_core/utilities/classes/@object_lookup/object_array.m
```diff
@@ -25,5 +26,20 @@
 end
 
 % Get return argument
+if numel(varargin)==1
+    iarray = varargin{1};
+    if ~isscalar(iarray)
+        error('Index to original object array, ''iarray'', must be a scalar')
+    end
+elseif numel(varargin)==0
```
oerc0122, line 34: Why not `isempty`?

herbert_core/utilities/classes/@object_lookup/object_array.m
```diff
@@ -25,5 +26,20 @@
 end
 
 % Get return argument
+if numel(varargin)==1
+    iarray = varargin{1};
+    if ~isscalar(iarray)
+        error('Index to original object array, ''iarray'', must be a scalar')
```
oerc0122, line 32: There is an ADR recommending an identifier for errors in the form of 
`'(HORACE|HERBERT):(function_name|ClassName):error_identifier'`
https://github.com/pace-neutrons/Horace/blob/master/documentation/adr/0023-error-format.md

oerc0122 (changes requested):

herbert_core/utilities/classes/@object_lookup/object_array.m
tgperring: Sheer force of habit. I think probably because line 29 tested if numel(varargin)==1 and this was the test if ==0. Does it make any real difference?

oerc0122: No, not really. Apparently it doesn't even flag as an mlint issue which I thought it would.

herbert_core/utilities/classes/@object_lookup/object_array.m
tgperring: Understood, and will follow the recommendations in future. I'd prefer not to edit this particular instance (and the possible one or two other new ones I introduced in this pull request) for the moment because since making this pull request Chris has asked me to reclone the repository following a bunch of changes that Alex, before I make any subsequent changes. This pull request would then straddle this process; I'd prefer to get this into master, and then edit later as there are going to be a lot of legacy error messages to edit in any case in the edited files. I also would like to clarify with you other aspects of the error handling as outlines in ADR 23 and 21.

PR #357, opened 2021-05-07
Fix outputs in simulate

oerc0122:
Fix `simulate`'s error handling (seems to be untested in Herbert?)

### DIFF
--- a/herbert_core/applications/multifit/@mfclass/simulate.m
+++ b/herbert_core/applications/multifit/@mfclass/simulate.m
@@ -96,7 +96,6 @@ function [data_out, calcdata] = simulate (obj, varargin)
 
 % Default return values if there is an error
 data_out = [];
-calcdata = [];
 
 % Check option
 opt_default = struct('sum',false,'foreground',false,'background',false,...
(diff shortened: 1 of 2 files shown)


herbert_core/applications/multifit/@mfclass/simulate.m
```diff
@@ -144,7 +144,7 @@
 end
 
 % Mask the data
-[wmask, msk_out, ok, mess] = mask_data_for_fit (obj.w_, obj.msk_);
+[wmask, msk_out] = mask_data_for_fit (obj.w_, obj.msk_);
```
cmarooney-stfc, line 147: Without the ok, mess returns on L.147, the next 3 lines won't work and have the potential to absorb a previously unset ok and/or an unset mess in the error_message.

cmarooney-stfc (changes requested):

abuts (approved):
good way of learning the code, but from the big picture point of view -- stylistic changes not worth much attention. Does unit test exist for simulation? If not, it would be more useful change.  It certainly missing for all identified issues.

cmarooney-stfc (approved):
Approving, note one comment previously sent

PR #360, opened 2021-05-24
Small routine to produce orthogonal vector and unit test for it

abuts:
This is small routine to produce two 3D vectors, orthogonal to the given one. Can be extended for more dimensions but not for the ticket

### DIFF
--- a/_test/test_utilities/test_ortho.m
+++ b/_test/test_utilities/test_ortho.m
@@ -1,0 +1,46 @@
+classdef test_ortho < TestCase
+    properties
+    end
+    methods
+        function obj = test_ortho(varargin)
+            if nargin<1
+                name = 'test_ortho';
+            else
+                name = varargin{1};
+            end
+            obj = obj@TestCase(name);
+        end
+        function test_ortho_throws(~)        
+            u = [0,0,0];
+            assertExceptionThrown(@()ortho_vec(u),'HERBERT:ortho_vec:invalid_argument')
+        end
+        
+        function test_ortho3(~)        
+            u = [1,-2,2];
+            [v1,v2] = ortho_vec(u);
+            
+            assertEqual(u*v1',0,'Vectors are not orthogonal',[1.e-9,0]);
+            assertEqual(u*v2',0,'Vectors are not orthogonal',[1.e-9,0]);            
+            assertEqual(v1*v2',0,'Vectors are not orthogonal',[1.e-9,0]);                        
+        end
+        
+        function test_ortho2(~)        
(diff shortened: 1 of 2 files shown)


_test/test_utilities/test_ortho.m
```diff
@@ -0,0 +1,46 @@
...
+                name = varargin{1};
+            end
+            obj = obj@TestCase(name);
+        end
+        function test_ortho_throws(~)        
+            u = [0,0,0];
+            assertExceptionThrown(@()ortho_vec(u),'HERBERT:ortho_vec:invalid_argument')
+        end
+        
+        function test_ortho3(~)
```
cmarooney-stfc, line 18: for all 3 tests, worth having more descriptive names. So (I think) test1 is a basic check that 2 orthogonal vectors are produced? Are 2 & 3 just doing the same for different input directions? Is there anything special about the choice of u in each case. Someone in the future is going to look at these tests, and think "why these particular ones?" Naming of the tests and comments will help reassure the maintainter what is expected here.

herbert_core/utilities/ortho_vec.m
```diff
@@ -0,0 +1,18 @@
...
+        'Input vector %s has zero size, so is orthogonal to any vector',...
+        evalc('disp(v0)'));
+end
+
+v2 = ones(size(v0));
+v2(ind) = 0;
+v1 = cross(v0,v2);
+v1 = v1/norm(v1);
+v2 = cross(v0,v1);
+v2 = v2/norm(v2);
```
cmarooney-stfc, line 16: Quite ingeneous. My equivalent solution for v2 is
        `[mvx,indx]=max(abs(v0))`
        `[mvn,indn]=min(abs(v0))`
        `v2 = ones(size(v0))`
        `v2(indx)=0`
        `v2(indn)=0`
        `indm=find(v2,1)`
        `v2 = v0`
        `v2(indx)=v0(indm)`
        `v2(indm)=-mvx`
which is a rotation of v0 about its smallest component direction. Yours is certainly more elegant.

cmarooney-stfc (approved):
Approved subject to a bit of commenting or naming on the tests

herbert_core/utilities/ortho_vec.m
abuts: interesting. Has advantage of selecting axis aligned orthogonal for axis aligned v0, but needs a lot of thinking to understand if works as intended. i.e. would not work for [1,1,1], though the idea is promising. I will leave my code as it is but feel free to improve appropriately

PR #375, opened 2021-07-31
fixing compilation on Ubuntu

abuts:
Minor changes, necessary for compilations on Ubuntu

### DIFF
--- a/admin/herbert_mex_mpi.m
+++ b/admin/herbert_mex_mpi.m
@@ -5,11 +5,11 @@ function [ok,mess] = herbert_mex_mpi(varargin)
 % Manually modify this script to specify the mpi libraries location in your
 % system.
 %
-use_her_mpich =true;
+use_her_mpich = true;
 if nargin > 0
-    verbouse = true;
+    verbose = true;
 else
-    verbouse = false;
+    verbose = false;
 end
 % the files, contributing into the communicator.
 input_files = {'cpp_communicator.cpp', 'input_parser.cpp', 'MPI_wrapper.cpp'};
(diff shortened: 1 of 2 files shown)


admin/herbert_mex_mpi.m
```diff
@@ -5,7 +5,7 @@
 % Manually modify this script to specify the mpi libraries location in your
 % system.
 %
-use_her_mpich = true
+use_her_mpich =true;
```
cmarooney-stfc, line 8: You have removed the balancing space to the right of the =. It would also be useful to fix the spelling of verbose from verbouse - always leaves the impression that perhaps a differently spelled instance may not be consistent.

cmake/herbert_FindMPI.cmake
```diff
@@ -2,9 +2,9 @@ set(EXTERNAL_ROOT "${Herbert_ROOT}/_LowLevelCode/external")
 if(UNIX)
     if (USE_HERBERT_MPI)
     	set(MPICH_VERSION "3.3a2")
-	set(MPI_ROOT "${EXTERNAL_ROOT}/glnxa64/mpich-${MPICH_VERSION}")
+	    set(MPI_ROOT "${EXTERNAL_ROOT}/glnxa64/mpich-${MPICH_VERSION}")
```
cmarooney-stfc, line 5: unclear why you are increasing the indent at this point

cmarooney-stfc (changes requested):
Minor typological changes but worth a fix while we're here

admin/herbert_mex_mpi.m
abuts: trailing space went off probably because I was switching it on/off and adding ; verbose done

cmake/herbert_FindMPI.cmake
abuts: that's how web interface interprets tabs. in fact,  no additional indent . I've fixed the tabs

PR #378, opened 2021-08-21
Fix init bug when no cluster worker

mducle:
In the initialisation step, there is a [call](https://github.com/pace-neutrons/Herbert/blob/rel_3_5/herbert_core/admin/%40opt_config_manager/opt_config_manager.m#L201) to load a saved configuration (in a local user folder). On first run, however, this file does not exist and instead a default configuration based on a guess of the system type (e.g. `win_small`, `win_large`, `idaas_small` etc). In some cases these defaults enable the MPI cluster to be on *but* the MPI worker `worker_v2.m` is no by default enabled on systems unless the user renames the `worker_v2.m.template`  with correct paths. 

Thus the initialisation fails with an error. Even though it errors, before this it *does* write a configuration file (with the cluster option set to false) and this is subsequently read and the initialisation then runs ok if you do a second `herbert_init`. This is fine (if a bit annoying) for the Matlab version, but for the Python (compiled Matlab) because `herbert_init` is called during the Matlab engine initialisation when this errors the Matlab engine is not deleted and you cannot call `herbert_init` again.

In any case having a lack of worker should not be an error - this is only an error in [`check_and_set_cluster_`](https://github.com/pace-neutrons/Herbert/blob/rel_3_5/herbert_core/classes/MPIFramework/%40parallel_config/private/check_and_set_cluster_.m) - in the similar [`check_and_set_worker_`](https://github.com/pace-neutrons/Herbert/blob/rel_3_5/herbert_core/classes/MPIFramework/%40parallel_config/private/check_and_set_worker_.m) function which is called before `check_and_set_cluster_`, it is a warning. However, the [load configuration routines](https://github.com/pace-neutrons/Herbert/blob/rel_3_5/herbert_core/configuration/%40config_base/config_base.m#L174) just loads from a structure defined for default machine types, and despite the worker field returning a warning which should prevent the cluster field being set, just goes ahead to try to set the cluster field.

This PR changes the cluster setter error to a warning instead.

An alternative implementation is to rework the load configuration code.

-----

To reproduce the bug, delete the existing configuration files (found using `herbert_config`) and run `herbert_init` again on a "large" machine (with more than 32GB memory) - you should see an error.

### DIFF
--- a/herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_cluster_.m
+++ b/herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_cluster_.m
@@ -11,8 +11,10 @@ wrkr = which(obj.worker_);
 mff = MPI_clusters_factory.instance();
 
 if isempty(wrkr)
-    error('HERBERT:parallel_config:not_available',...
+    warning('HERBERT:parallel_config:not_available',...
         'Parallel worker is not on the Matlab path so parallel features are not available')
+    config_store.instance().store_config(obj,...
+        'parallel_cluster','none','cluster_config','none');
 else
     known_clusters = mff.known_cluster_names;
     full_cl_name = obj.select_option(known_clusters,cluster_name);


herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_cluster_.m
```diff
@@ -11,8 +11,10 @@
 mff = MPI_clusters_factory.instance();
 
 if isempty(wrkr)
-    error('HERBERT:parallel_config:not_available',...
+    warning('HERBERT:parallel_config:not_available',...
         'Parallel worker is not on the Matlab path so parallel features are not available')
+    config_store.instance().store_config(obj,...
+        'parallel_cluster','none','cluster_config','none');
```
cmarooney-stfc, line 17: @mducle @abuts Hi Duc, Thanks for all the diagnostic work you put in here. I am wondering if this is the best approach. Horace emits warnings left right and centre and it is easy to ignore them. What will the state of execution be if the warning is emitted and the code carries on? My feeling is that there should be 2 possibilities. If the user has not selected to run parallel, then a warning should be acceptable. But if the user has selected parallel then the run should error. Ideally the initialisation should not be wrong in the first place. What do you think?

mducle: Sorry... forgot about this PR.

> What will the state of execution be if the warning is emitted and the code carries on?

It will proceed without using the cluster (e.g. with only shared-memory / thread parallelisation).

The main issue here is the hard-coding on the types of "machines" (e.g. `win_large` / `idaaas_large`) to automatically using clustering even though the cluster worker is not defined - the user needs to set-up paths for the cluster workers so this cannot be automatic...

If we don't put the check here then we should put it in the code which detects which type of machine it is / or the code which sets up what options each type of machine - to check if the worker script has been set up...

I was really hoping that @abuts would sort this out really...

mducle:
@cmarooney-stfc 

Actually... just had a chat with Alex, and the [`horace_install`](https://github.com/pace-neutrons/Horace/blob/rel_3_5/admin/horace_install.m) script actually creates `worker_v2.m` from `worker_v2.m.template`... - this is what users should have if they [follow the instructions](https://pace-neutrons.github.io/Horace/unstable/Download_and_setup.html#installation-instructions).

However, I wasn't doing this because I wanted to switch between versions and was using just `addpath('herbert_core'); herbert_init; addpath('horace_core'); horace_init` instead of using `horace_install`.

Anyway, we decided that it should probably stay an error... whether to automatically enable the cluster parallelisation for different types of machines I'll leave to another issue... (if somebody else complains ;)...

Instead, I'll implement a work-around for `pace-python`.

cmarooney-stfc (commented):
@mducle @abut I've just reverted this to Review/QA for 3 reasons.
(1) Purely administrative - when complete please set to Done (not Closed) for all Horace and Herbert tickets - this leaves it visible for the next sprint review
(2) You say in the next comment - "it should stay an error ...." but it hasn't, it's become a warning. Happy for it to be an error but it hasn't stayed that way.
(3) Please could you create the ticket for automatically enabling cluster parallelisation for different machines. Reference this ticket and your work for pace-python.
thanks - please set to done when you're finished.

mducle:
@cmarooney-stfc - I wanted to close this without merging - e.g. no changes to the code, hence why I didn't think it should be "done" and why it stays an error.

cmarooney-stfc:
@mducle Hi Duc thanks for the clarification. I'll leave the PR as done so it is easily visible for the next sprint and then I'll close it. I hadn't appreciated that it was still error as no merge took place

Do you think you can do the new ticket on auto enabling cluster parallelization? You have done the consultation with Alex so I think it would be clearer from you.
thanks Chris

PR #385, opened 2021-09-29
New sqw new instrument classes

cmarooney-stfc:
Herbert infrastructure for new experiment classes which reorganize the sqw header to provde more flexible instrument and sample


herbert_core/classes/instrument_classes/instrument/instruments/@IX_null_inst/IX_null_inst.m
```diff
@@ -0,0 +1,270 @@
...
+            %makes the null struct for storage in a file
+            str = struct();
+        end
+    end
+
+    %======================================================================
+    % Methods for fast construction of structure with independent properties
+    methods (Static, Access = private)
+        function names = propNamesIndep_
+            % Determine the independent property names and cache the result.
```
abuts, line 28:
```suggestion
            % Determine the independent property names and cache the result.
            persistent names_store
            if isempty(names_store)
                names_store = fieldnamesIndep(eval(mfilename('class')));
            end
            names = names_store;            
```
this is private function specific to the class. eval('somethin') is usually least effective way of doing things and very difficult to debug. 
Should be: -- much simpler and efficient
```
 names_store = fieldnamesIndep(IX_null_inst());
```

herbert_core/classes/instrument_classes/instrument/@IX_null_sample/IX_null_sample.m
```diff
@@ -9,12 +9,258 @@
...
+            %
+            % Input:
+            % ------
+            %   obj     Scalar instance of the object class
+            %
+            % Output:
+            % -------
+            %   S       Structure created from obj that is to be saved
+            
+            % The following is boilerplate code
```
abuts, line 225: To avoid issues with changes in a future, the objects should be saved with version number.
not sure what is better -- specify constant property `version =1 `  (will `structIndep` pick it up?)
or add here
```
S.version =1
```
loadobj should strip the version accordingly (and process data according to version)

abuts (changes requested):
All these needs unit tests.  -- this is the main issue -- the code is not verified at all. 

e.g. I am not sure if, say loadob in this form will work for array of objects (it looks like it will return cellarray of appropriate objects. A test would immediately prove if I am wrong or right.

All classes should have class_version_ 
probably constant private property (if inheritance expected)
if `fieldsIndep`  would return this property, then probably OK, if not, `saveobj` should add it to the structure to save. 
The question remains about efficient load/save array of objects. These should certainly be converted using `struct2cell` & `cell2struct` operations -- the performance would be dreadful otherwise. 

All other things are more or less ok in the frames of our discussion.

cmarooney-stfc:
Merging before creating the tests to allow a Horace PR to use these changes in master. Must finish tests as soon as that PR passes

PR #387, opened 2021-10-04
structure comparison using all(ismember()) operation

abuts:
trivial fix, as described in ticket

### DIFF
--- a/herbert_core/utilities/maths/equal_to_tol.m
+++ b/herbert_core/utilities/maths/equal_to_tol.m
@@ -104,9 +104,7 @@ function [ok,mess]=equal_to_tol(a,b,varargin)
 
 % Original author: T.G.Perring
 %
-% $Revision:: 840 ($Date:: 2020-02-10 16:05:56 +0000 (Mon, 10 Feb 2020) $)
-
-
+%
 % The following code is pretty complex as it has to handle legacy input as
 % well. Touch at your peril!
 warn = warning('off','MATLAB:structOnObject');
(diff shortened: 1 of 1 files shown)


herbert_core/utilities/maths/equal_to_tol.m
```diff
@@ -405,7 +405,8 @@
 
     % Check fieldnames are identical
     fields=fieldnames(a);
-    if ~isequal(fields,fieldnames(b))
+    fb = fieldnames(b);
```
oerc0122, line 408: Would prefer `fieldsA`, `fieldsB` for clarity and consistency

herbert_core/utilities/maths/equal_to_tol.m
```diff
@@ -405,7 +405,8 @@
...
+    if ~(isempty(fields)&&isempty(fb)) && ~any(ismember(fields,fb))
         ok=false;
```
oerc0122, line 410: Would use this as an opportunity to use error failure rather than returning `ok`

oerc0122 (changes requested):

herbert_core/utilities/maths/equal_to_tol.m
abuts: its used in number of places with ok check (including Horace) I would be highly scared to change all of these. Its also possible that this generic function is not creating errors if the objects are different but uses it in some logic. 

I suspect, changing it to throwing an error would bring much more work then one would like to accept.

cmarooney-stfc (approved):

PR #392, opened 2021-10-14
Deprecate multifit_legacy temporarily

oerc0122:
Should be able to permanently remove multifit/tobyfit legacy when we can prove that it does not affect anything.


abuts (approved):
looks good as soon as tests are disabled too. 
I think, removing them from cmake list would be sufficient

PR #399, opened 2021-11-11
dlpl issue

abuts:
apparently its a feature, that Matlab plots only continuous lines if data are interrupted by NaN-s. 
If such situation happens, let's issue a warning informing user that he is not seeing number of points. 

To test:
go to  babylon5\Public\abuts\plot_bug\ and run script which loads data with Nans and demonstrate that line is not plotted. 
Switch to the branch and see warning message, explaining why data are not plotted.


herbert_core/graphics/@IX_dataset_1d/private/plot_oned.m
```diff
@@ -149,6 +149,13 @@
 elseif plot_type(1)=='h'
     plot_histogram (w);
 elseif plot_type(1)=='l'
+    [frac, np] = w.calc_continuous_fraction();
+    if frac<0.5
+        warning('HERBERT:IX_dataset_1D:invalid_argument',...
+            [' Your data contain NaN-s and you are plotting %3.1f%%',...
```
cmarooney-stfc, line 155: data containS. Appreciating that data can sometimes be plural, but here it is a singular container

cmarooney-stfc (approved):

PR #400, opened 2021-11-14
Ensure no empty struct instruments or samples

cmarooney-stfc:
There have been 2 alternative paradigms used for instruments and samples that were not defined. One was to use an empty class object (with isempty redefined to check no fields used) and one was to use IX_null versions of the instrument and sample classes if not definition was provided. This alteration enables the latter paradigm, and also makes the base classes IX_inst and IX_samp abstract to prevent their use for creating objects. The IX_null classes have also been made (in Alex' sense) serializable.

Additional changes have been made to make the tests run.

This may break Horace which has not yet been updated to fit.

### DIFF
--- a/herbert_core/classes/data_loaders/@rundata/rundata.m
+++ b/herbert_core/classes/data_loaders/@rundata/rundata.m
@@ -73,9 +73,9 @@ classdef rundata
         oriented_lattice_ =[];
         
         % instrument model holder;
-        instrument_ = IX_inst();
+        instrument_ = IX_null_inst();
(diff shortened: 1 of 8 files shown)


abuts (commented):
can not review carefully, but my feeling is that loadobj has to be overloaded for each child and the IX_null_sample and IX_nul_inst should have isempty overloaded for true, (can be removed from all other classes)

cmarooney-stfc:
This has been obsoleted by other developments and there is now no point in keeping it

PR #401, opened 2021-11-15
Brief improvements to parallel framework

oerc0122:
Complete set of convenience functions.

Add a `setup` phase to job executor called once before entering calculation loop (non-abstract to avoid having to modify all existing JobExecutors with null function)

### DIFF
--- a/herbert_core/classes/MPIFramework/@JobExecutor/JobExecutor.m
+++ b/herbert_core/classes/MPIFramework/@JobExecutor/JobExecutor.m
@@ -20,6 +20,7 @@ classdef JobExecutor
     % JobExecutor Properties:
     %
     % labIndex          - The number of the running task.
+    % numLabs           - The number of running tasks.
     % mess_framework    - Access to messages framework used for messages
     %                     exchange between the parallel tasks.
     % control_node_exch - The instance of the framework used to exchange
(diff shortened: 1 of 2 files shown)


herbert_core/classes/MPIFramework/@JobExecutor/JobExecutor.m
```diff
@@ -20,6 +21,7 @@
     % JobExecutor Properties:
     %
     % labIndex          - The number of the running task.
+    % numLabs           - The number of running tasks.
```
cmarooney-stfc, line 24: Just to say it could be good to say why "lab" is used here. Presuming it means label? taskIndex etc look more intuitive

herbert_core/classes/MPIFramework/@JobExecutor/JobExecutor.m
```diff
@@ -343,6 +359,18 @@ function log_progress(this,step,n_steps,time_per_step,add_info)
             [ok,err] = obj.mess_framework_.send_message(message);
         end
 
+        function [ok,err,message]=receive_message(obj,from_task_id,varargin)
+            % Wrapper to receive_message to ease writing
+            [ok,err,message] = obj.mess_framework_.receive_message(from_task_id,varargin{:});
+        end
+
+        function obj = setup(obj)
+            % Function called one before entering do_job loop
```
cmarooney-stfc, line 368: Not clear what "one before" means. Did you just mean "before"?

herbert_core/classes/MPIFramework/parallel_worker.m
```diff
@@ -213,22 +213,23 @@
...
+
+
         % send first "running" log message and set-up starting time. Runs
         % asynchronously.
         n_steps = je.n_steps;
         if DO_LOGGING; log_disp_message('Logging start and checking for job cancellation before loop je.is_completed loop\n'); end
         mis.do_logging(0,n_steps);
         %%
-        
+        je = je.setup()
```
cmarooney-stfc, line 226: I know you've documented setup in its place for je, but it would help to say here what you're doing so it is as documented as the rest of the workflow here

cmarooney-stfc (changes requested):

herbert_core/classes/MPIFramework/@JobExecutor/JobExecutor.m
abuts: used Matlab convention for openMPI interface, which calls independent MPI worker(s) -- lab(s)

PR #403, opened 2021-12-07
Create utility functions to be used as part of parallel multifit test…

tgperring:
…s yet to be created. Fix a couple of bugs in Gaussian fit functions for 4D data sets.

Also edited IX_dataset noisify methods to correctly compute error bars. The error property was being filled with variances not standard deviations until now. This may have been introduced in late 202 with the refactoring of the utility function utility\maths\noisify.


herbert_core/applications/mfit_funcs/slow_func1d.m
```diff
@@ -0,0 +1,44 @@
+function y = slow_func1d (x, p, funchandle, nslow)
```
oerc0122, line 1: Rather than defining this function multiple times, an alternative would be to pass `x` in as a cell array and use cellarray expansion to fill the funchandle.

herbert_core/applications/multifit/_test/data/create_testdata_multifit_1.m
```diff
@@ -24,29 +24,29 @@
...
 [y3,e3]=noisify(y3,e3);
-c3 = {x3,y3,e3};
-s3=struct('x',x3,'y',y3,'e',e3);
-w3=IX_dataset_1d(x3,y3,e3,'Test Gaussian 3','x-axis','signal');
+c3 = {x3,y3,sqrt(e3)};
+s3=struct('x',x3,'y',y3,'e',sqrt(e3));
+w3=IX_dataset_1d(x3,y3,sqrt(e3),'Test Gaussian 3','x-axis','signal');
 
 save(output_file,...
     'p1','x1','y1','e1','c1','s1','w1',...
```
oerc0122, line 52: Saving not-sqrt'd vals.

herbert_core/utilities/general/time_waster.m
```diff
@@ -0,0 +1,41 @@
...
+% ------
+%   yin     Array of values
+%   nloop   Number of times to perform the operation
+%
+% Output:
+% -------
+%   yout    Output array
+
+
+if nloop~=0
```
oerc0122, line 24: Don't strictly need to check here, 
```matlab
for i=1:0
 ...
end
```
Should not execute. It may, however, spend some time doing the elementwise product, I guess.

herbert_core/applications/mfit_funcs/slow_func1d.m
```diff
@@ -0,0 +1,44 @@
...
+    nslow = 1;
+else
+    nslow = round(nslow);
+    if nslow < 0
+        nslow = 0;
+    end
+end
+y = funchandle (x, p);
+
+if nslow > 0
```
oerc0122, line 42: We don't need to protect against this, `time_waster` already does.

herbert_core/applications/mfit_funcs/slow_func1d.m
```diff
@@ -0,0 +1,44 @@
...
+%               Each value of nslow takes about the time of 25 
+%               exponentiations.
+%                   nslow >=0
+%               Default if not given: nslow = 1
+%
+% Output:
+% ========
+%   y           Array of calculated y-axis values
+
+if nargin==3
```
oerc0122, line 32: Also, rather than using the less clear `nargin` approach it might be better to use an inputParser. Cost is negligible as this is time-wasting. 
```matlab
ip = inputParser();
addRequired(ip, 'x', @iscell);
addRequired(ip, 'p', @(x) validateattributes(x, {'numeric'}, {'vector'}}));
addRequired(ip, 'funchandle', @(x) isa(x, 'function_handle'));
addOptional(ip, 'nslow', 1, @(x) validateattributes(x, {'numeric'}, {'scalar', 'integer', 'nonnegative'}));
parse(ip, x, p, funchandle, nslow);
[x, p, funchandle, nslow] = struct2cell(ip.Results);
```

herbert_core/applications/mfit_funcs/slow_func1d.m
```diff
@@ -0,0 +1,44 @@
...
+    nslow = 1;
+else
+    nslow = round(nslow);
+    if nslow < 0
+        nslow = 0;
+    end
+end
+y = funchandle (x, p);
```
oerc0122, line 40: If you choose to rewrite this taking `x` as a cellarray, this becomes
```matlab
y = funchandle(x{:}, p);
```

herbert_core/applications/multifit/_test/data/create_testdata_multifit_1.m
```diff
@@ -24,29 +24,29 @@
 x1=-10:2:90;
 x1=x1 + 0.1*rand(size(x1));
 y1=mftest_gauss_bkgd(x1,p1);
-e1=sqrt(y1);
+e1=y1;
```
oerc0122, line 27: As noted previously, perhaps rewrite as
```matlab
[y1, e1] = noisify(y1, y1);
e1 = sqrt(e1);
...
```

herbert_core/utilities/classes/@IX_data_2d/noisify.m
```diff
@@ -16,5 +16,7 @@
 %
 wout=w;
 for i=1:numel(w)
-    [wout(i).signal,wout(i).error]=noisify(w(i).signal,w(i).error,varargin{:});
+    [wout(i).signal, errsqr] = noisify (...
```
oerc0122, line 19: Rather than maintaining N copies of this, add this to the IX_dataset base-class?

oerc0122 (changes requested):
Could be simplified down a bit, though mostly cosmetic.

herbert_core/applications/mfit_funcs/slow_func1d.m
oerc0122: **N.B.** I have not checked this works, it is just an example.

_test/test_multifit/create_testdata_multifit_1.m
```diff
@@ -22,23 +22,23 @@
 
 x1=0:2:100 + 0.1*rand(1,100);
 y1=mftest_gauss_bkgd(x1,p1);
-e1=sqrt(y1);
+e1=y1;
 [y1,e1]=noisify(y1,e1);
-wstruct1=struct('x',x1,'y',y1,'e',e1);
-w1=IX_dataset_1d(x1,y1,e1,'Test Gaussian 1','x-axis','signal');
```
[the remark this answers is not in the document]
tgperring: It was the minimum perturbation to the code. Given that optimisation is unimportant here (they are tiny datasets) I thought it best to make the change as small as possible.

_test/test_multifit/create_testdata_multifit_1.m
```diff
@@ -22,23 +22,23 @@
...
 
 x3=0:2:100 + 0.1*rand(1,100);
 y3=mftest_gauss_bkgd(x3,p3);
-e3=sqrt(y3);
-[y3,e3]=noisify(y3,e3);
-wstruct3=struct('x',x3,'y',y3,'e',e3);
+e3=y3;
+[y3,e3]=noisify(y3,sqrt(e3));
+wstruct3=struct('x',x3,'y',y3,'e',sqrt(e3));
 w3=IX_dataset_1d(x3,y3,e3,'Test Gaussian 3','x-axis','signal');
```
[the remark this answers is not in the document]
tgperring: This is an error on my part. I'll correct it.

_test/test_multifit/create_testdata_multifit_1.m
```diff
@@ -22,23 +22,23 @@
...
 
 save(output_file,'p1','x1','y1','e1','wstruct1','w1','p2','x2','y2','e2','wstruct2','w2','p3','x3','y3','e3','wstruct3','w3')
```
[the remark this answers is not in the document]
tgperring: No, tracing through the tests further downstream this is not what is wanted as e1, e2, e3 as the standard deviations is what is required.

_test/test_multifit_parallel/EXAMPLE_test_script_1.m
```diff
@@ -0,0 +1,138 @@
+%% ===================================================================================
```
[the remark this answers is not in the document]
tgperring: Sounds like a good idea. My intention was that you would use the examples as you saw fit, but a more general example_codes folder seems valuable, so long as we do regular house-keeping.

herbert_core/applications/mfit_funcs/slow_func1d.m
tgperring: Just never used it before, and wanted to get something in quick...

herbert_core/applications/multifit/_test/data/create_testdata_multifit_1.m
tgperring: Noted. BTW, these tests are somewhat flakey; they are on my to-do list to overhaul as they are not automatically performed, and in the case of this function (for example) there is duplication of the same file as the one in herbert/_test/test_multifit.

herbert_core/utilities/classes/@IX_data_2d/noisify.m
tgperring: In a parallel branch I am doing a complete overhaul of IX_dataset and all its children, with one of the goals to replace the 1D,2D,3D versions of a method with just one generic n-dimensional method; there are many such methods, not just noisify. here my intention is just to get the noisify methods to work correctly as swiftly as possible.

_test/test_multifit/create_testdata_multifit_1.m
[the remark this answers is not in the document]
tgperring: However, the error I made in not noticing that the call to save is now saving different data (see reply to a comment below) makes this redundant as I will follow your advice on code simplification below.

_test/test_multifit/create_testdata_multifit_1.m
```diff
@@ -22,23 +22,23 @@
 
 x1=0:2:100 + 0.1*rand(1,100);
 y1=mftest_gauss_bkgd(x1,p1);
-e1=sqrt(y1);
+e1=y1;
 [y1,e1]=noisify(y1,e1);
```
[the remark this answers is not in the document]
tgperring: Indeed. See my reply to an earlier comment.

herbert_core/applications/mfit_funcs/slow_func1d.m
tgperring: I'm making the check in time_waster more robust so I can eliminate the check in slow_func1d and 2d, 3d friends

herbert_core/applications/mfit_funcs/slow_func1d.m
tgperring: The multifit API for user functions for different dimensionalities pass different numbers of arguments: (x1... for 1D, (x1,x2,... for 2D etc. which is why there are separate functions slow_func1d, slow_func2d etc. I could make a single function 
           function slow_func (varargin)    
by checking the position of the function handle in the input argument list, but that would make argumnet checking less robust.

herbert_core/utilities/general/time_waster.m
```diff
@@ -14,25 +14,30 @@
...
+%   nloop   Number of times to perform the operation. If nloop <= 0 then
+%           the operation is ignored.
 %
 % Output:
 % -------
 %   yout    Output array
 
 
-if nloop~=0
+if round(nloop)>0
```
oerc0122, line 25: Why bother rounding? It's visual noise in the function. In matlab
```matlab
for j = 1:6.6
   j
end
% 1 2 3 4 5 6
```
If someone's silly enough to pass in a fractional number to what clearly should be an int (through name alone), that being undefined (or round down in this case) is fine.

 If you're still concerned, round it once at the beginning of the function and keep the rest of the function clean.

oerc0122 (approved):
If it's all working, that looks mostly fine apart from my one note

PR #405, opened 2022-01-06
loop if more than one plot

cmarooney-stfc:
For IX_data_1d, the function calc_continuous_fraction as implemented in the private function calc_cont_frac_, scans a single plot for points which cannot be plotted because although their value is plottable, the values either side are NaN and so lines cannot be drawn from the plottable point.

In the tests in test_tobyfit, calls to this function with one such plot work fine. However there are IX_data_1d objects with three plots in them, and they fail because the first line of calc_cont_frac_ fails - ~valid has too many arguments (it says). This PR fixes the behaviour.

As modified here the parent call calc_continuous_fraction now loops over the component plots, returns the corresponding fraction, and reports to the caller the minimum fraction of the set. This minimizes the changes required to run the tests.

### DIFF
--- a/herbert_core/utilities/classes/@IX_data_1d/IX_data_1d.m
+++ b/herbert_core/utilities/classes/@IX_data_1d/IX_data_1d.m
@@ -113,18 +113,9 @@ classdef IX_data_1d < IX_dataset
                       ['array of IX_data_1d has more than 2 dimensions, ' ...
                       'and cannot be used with calc_continuous_fraction']);
             end
-            frac = +2.0; % >1 so enough to trigger fr<frac on any pass below
-            n_points = 0;
-            sz = size(obj);
-            for i = 1:sz(1)
-                for j = 1:sz(2)
-                    [fr,np] = calc_cont_frac_(obj(i,j));
-                    if fr<frac
-                        frac = fr;
-                        n_points = np;
-                    end
-                end
-            end
+            [fr,np] = arrayfun(@calc_cont_frac_, obj);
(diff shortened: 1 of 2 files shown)


herbert_core/utilities/classes/@IX_data_1d/IX_data_1d.m
```diff
@@ -95,13 +95,36 @@
...
+            nd = ndims(obj);
+            if nd>2
+                error('HERBERT:IX_data_1d:calc_continuous_fraction', ...
+                      ['array of IX_data_1d has more than 2 dimensions, ' ...
+                      'and cannot be used with calc_continuous_fraction']);
+            end
+            frac = +2.0; % >1 so enough to trigger fr<frac on any pass below
+            n_points = 0;
+            sz = size(obj);
+            for i = 1:sz(1)
```
oerc0122, line 119: In my understanding of MATLAB optimisation: MATLAB loops are bad, MATLAB loops with ifs are double bad. 

This will probably be faster, but more memory intensive (if that's a problem).
```matlab
[fr,np] = arrayfun(@calc_cont_frac_, obj);
[frac, ind] = min(fr);
n_points = np(ind);
```

In my eye, it's a bit cleaner too, the entire function reduces to 5 clear lines.

cmarooney-stfc: Nice one! adopted

cmarooney-stfc:
Merging as reviewed, 3 ok builds, failed build due to MPI ie not related to this PR.

PR #406, opened 2022-01-11
Update docify

mducle:
* Removed docify-generated docstrings from all functions.
* Add overloaded `help` and `doc` functions which run docify and return the outputs if needed on-the-fly.
* Add a `cmake` command to run docify as part of the build process.

In principle the overloaded `help` and `doc` commands should generate the same help text as if the docstrings were actually there, but there are some caveats:

* `help` and `doc` for methods and properties defined in the same file as the main classdef are not docified-on-the-fly (methods defined in separate files are fine). However, it seems that docify does not support replacing descriptions of _properties_ within the main classdef file in any case (an example is `help mfclass/fun`), and there seems to be no case of docified docstrings in methods in classdef files.
* `help` and `doc` for inherited class methods are not docified-on-the-fly. An example is `help IX_data_1d/acosh` because `IX_data_1d` inherits from `IX_dataset` - `help IX_dataset/acosh` works as expected.
* `doc` summary in the main class page for methods are not docified (only the main page is run using `docify_single`)

As a result of these limitations, the overloaded functions should not be used for release versions but are convenience functions for developers so we can remove the ~10,000 lines of repeated docstrings in the code. For the release versions, the build process now runs `docify` to generate the docstrings before packaging.

(this is an update of PR #404 - redone as two commits, [one](https://github.com/pace-neutrons/Herbert/commit/c3c2b31a2823388e188a869bd413f3b731b74fda) with just the code changes; the [other](https://github.com/pace-neutrons/Herbert/commit/846eeae1b88b1a4cb9290347bcfc714e30237880) with all the docstrings removed)


cmarooney-stfc (commented):
@mducle Nothing specific to comment on in these code changes, but a general worry.

In the description in docify/doc.m and in the PR description you say that there are general problems with the documentation not grabbing the correct descriptions for some properties etc. Overall this gives the impression that maybe we should think again about doing this? Or at least that the description should provide the pros and cons? Or pointing out that the original had equivalent problems?

@tgperring Just a jog that you are also down as a reviewer

PR #407, opened 2022-01-12
Use null experiment components

cmarooney-stfc:
IX_inst and IX_samp, being the base classes of IX_inst_DGfermi (and other instruments) and IX_sample, have been deprecated as classes representing absence of instrument or sample data. IX_null_inst and IX_null_sample are used instead. This PR makes the change.

I am also suspect that `from_old_structure` function should be defined to fix the tests as there were some tests for `saveobj` in old form. If not then not....


cmarooney-stfc:
This PR parallels an equivalent one about to be made in Horace. Individual PRs should fail, they have been tested together in a branch build and are OK.

herbert_core/classes/instrument_classes/instrument/@IX_samp/IX_samp.m
```diff
@@ -331,6 +345,7 @@
     
     methods
         %------------------------------------------------------------------
+        %{
```
abuts, line 348: If you make class serializable, this method should be removed

herbert_core/classes/instrument_classes/instrument/@IX_samp/IX_samp.m
```diff
@@ -349,10 +364,12 @@
             
             S = structIndep(obj);
         end
+        %}
     end
     
     %------------------------------------------------------------------
     methods (Static)
+        %{
```
abuts, line 372: If you make class serializable, this method should be modified according to template.
                obj = IX_samp();
                obj = loadobj@serializable(S,obj);

as there is empty constructor involved, the class can not be abstract

herbert_core/classes/instrument_classes/instrument/instruments/@IX_inst/IX_inst.m
```diff
@@ -287,6 +299,7 @@
         end
     end
     methods(Sealed)
+        %{
         function is = isempty(obj)
```
abuts, line 303: as you introduced null_inst, this either should not here or always return false (or array of false, the size of obj -- certainly when heterogeneous is removed) 
You probably better remove this method, as default `isempty` would return `false` so no point of creating duplication. 

null_inst should overload `isempty` to `true`

herbert_core/utilities/classes/@IX_data_1d/IX_data_1d.m.orig
```diff
@@ -0,0 +1,158 @@
+classdef IX_data_1d < IX_dataset
+    % IX_data_1d class implements main operation with 1-dimensional data
```
abuts, line 2: this file should be removed (`.orig`)

abuts (changes requested):
partial remarks.

herbert_core/utilities/classes/@IX_data_1d/IX_data_1d.m.orig
cmarooney-stfc: done

PR #408, opened 2022-01-14
Update FindMatlab.cmake

oerc0122:
Add new MATLAB versions


abuts (approved):
I did it in the past -- this is not enough for finding new versions. It does something more with the versions. But the change makes no harm.

cmarooney-stfc (approved):
Presuming that this is all it takes to make these Matlabs available, and in that case, approving

PR #409, opened 2022-01-14
763 fix herb exc 2021a

oerc0122:


abuts (approved):
looks ok. I hope new Matlab also returns exception level.

PR #434, opened 2022-03-20
Compiled workers - Herbert changes

mducle:
Changes the checks in `@parallel_config` and `@MPI*` classes to allow compiled workers.

When the `worker` field of `parallel_config` is set to something which is _not_ a Matlab function, then it is checked to see if it may be a compiled worker. This check considers something a compiled worker if:

* It is a full path to a existing file; OR
* It is a command on the system `PATH` (as determined by the `which` (Linux) or `where` (Windows) command).

The full path to the file is set as the worker and the `is_compiled` flag is set to true. No check is made as to whether this file is actually executable or can run Horace etc... if it is not then any parallel computation will fail further down the chain; such checks would involve (perhaps significant) modification of `worker_v2` and/or setting up an actual computation which would take too long... we *assume* users would not set a worker they know will not work... (and if they do they will get an error when they try to run a computation...)

The actual compiled worker code is in a [Horace PR #794](https://github.com/pace-neutrons/Horace/pull/794). Tests will be added in the Horace PR (later!)

### DIFF
--- a/herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
+++ b/herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
@@ -173,14 +173,13 @@ classdef ClusterMPI < ClusterWrapper
             external_dll_dir = fullfile(rootpath, 'DLL','external');
             if ispc()
                 [rs, rv] = system('where mpiexec');
-                if rs == 0
-                    mpi_exec = strip(rv);
-                    mpis = splitlines(mpi_exec);
-                    if numel(mpis) > 1
-                        % If multiple mpiexec on path, prefer user installed MS MPI
-                        mpi_id = [1 find(cellfun(@(x) ~isempty(strfind(x,'Microsoft')), mpis), 1)];
-                        mpi_exec = mpis{max(mpi_id)};
-                    end
+                mpis = splitlines(strip(rv));
+                % Ignore Matlab-bundled mpiexec (firewall issues)
+                mpis(cellfun(@(x) contains(x, matlabroot), mpis)) = [];
+                if rs == 0 && ~isempty(mpis)
+                    % If multiple mpiexec on path, prefer user installed MS MPI
+                    mpi_id = [1 find(cellfun(@(x) contains(x,'Microsoft'), mpis), 1)];
+                    mpi_exec = mpis{max(mpi_id)};
                 else
                     % No mpiexec on path, use pre-packaged version
                     mpi_exec = fullfile(external_dll_dir, 'mpiexec.exe');
--- a/herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
+++ b/herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
@@ -178,9 +178,8 @@ classdef ClusterMPI < ClusterWrapper
                     mpis = splitlines(mpi_exec);
                     if numel(mpis) > 1
                         % If multiple mpiexec on path, prefer user installed MS MPI
-                        mpi_id = find(cellfun(@(x) ~isempty(strfind(x,'Microsoft')), mpis), 1);
-                        if isempty(mpi_id), mpi_id = 1; end
-                        mpi_exec = mpis{mpi_id};
+                        mpi_id = [1 find(cellfun(@(x) ~isempty(strfind(x,'Microsoft')), mpis), 1)];
+                        mpi_exec = mpis{max(mpi_id)};
                     end
                 else
                     % No mpiexec on path, use pre-packaged version
--- a/herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
+++ b/herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
@@ -172,9 +172,20 @@ classdef ClusterMPI < ClusterWrapper
             rootpath = fileparts(which('herbert_init'));
             external_dll_dir = fullfile(rootpath, 'DLL','external');
             if ispc()
-                % only one version of mpiexec is used now. May change in the
-                % future.
-                mpi_exec = fullfile(external_dll_dir, 'mpiexec.exe');
+                [rs, rv] = system('where mpiexec');
+                if rs == 0
+                    mpi_exec = strip(rv);
+                    mpis = splitlines(mpi_exec);
+                    if numel(mpis) > 1
+                        % If multiple mpiexec on path, prefer user installed MS MPI
+                        mpi_id = find(cellfun(@(x) ~isempty(strfind(x,'Microsoft')), mpis), 1);
+                        if isempty(mpi_id), mpi_id = 1; end
+                        mpi_exec = mpis{mpi_id};
+                    end
+                else
+                    % No mpiexec on path, use pre-packaged version
+                    mpi_exec = fullfile(external_dll_dir, 'mpiexec.exe');
+                end
             else
                 mpi_exec = fullfile(external_dll_dir, 'mpiexec');
                 
--- a/herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
+++ b/herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
@@ -94,6 +94,9 @@ classdef ClusterParpoolWrapper < ClusterWrapper
             end
             
             obj = init@ClusterWrapper(obj,n_workers,mess_exchange_framework,log_level);
+            assert(~obj.is_compiled_script_, ...
+                'HERBERT:ClusterParpoolWrapper:invalid_argument', ...
+                'Parpool cluster does not work with compiled workers')
             
             % delete interactive parallel cluster if any exist
             cl = gcp('nocreate');
@@ -137,11 +140,7 @@ classdef ClusterParpoolWrapper < ClusterWrapper
             % variables, but if the cluster is remote, the envriomental
             % variables transfer should be investigated
             obj.set_env();
-            if obj.is_compiled_script_
-                h_worker = @worker_v2;
-            else
-                h_worker = str2func(obj.worker_name_);
-            end
+            h_worker = str2func(obj.worker_name_);
             task = createTask(cjob,h_worker,0,{cs});
             
             obj.cluster_ = cl;
--- a/herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
+++ b/herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
@@ -137,7 +137,11 @@ classdef ClusterParpoolWrapper < ClusterWrapper
             % variables, but if the cluster is remote, the envriomental
             % variables transfer should be investigated
             obj.set_env();
-            h_worker = str2func(obj.worker_name_);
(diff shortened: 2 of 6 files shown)


herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_cluster_.m
```diff
@@ -7,13 +7,11 @@
 % The cluster name (can be defined by single symbol)
 % or by a cluster number in the list of clusters
 %
-wrkr = which(obj.worker_);
-mff = MPI_clusters_factory.instance();
+    assert(~isempty(which(obj.worker)) || exist(obj.worker, 'file'), ...
```
oerc0122, line 10: This code seems duplicated 3x is it worth making it a function?

herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_worker_.m
```diff
@@ -32,6 +40,23 @@
...
 else % worker function is available.
-    config_store.instance().store_config(obj,'worker',new_wrkr);
+    config_store.instance().store_config(obj, 'worker', new_wrkr);
+    config_store.instance().store_config(obj, 'is_compiled', false);
 end
+end % function
 
+function out = check_compiled_(worker)
+    out = '';
+    if exist(worker, 'file')
```
oerc0122, line 50: Use `is_file(worker)` (`Herbert/herbert_core/utilities/general/is_file.m`)

herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_worker_.m
```diff
@@ -32,6 +40,23 @@
...
+        % Assume if input is full path to file, then it is a compiled worker
+        out = worker;
+    else
+        if ispc(), cmd = 'where'; else, cmd = 'which'; end
```
oerc0122, line 54: I personally dislike inline-`if`s like this. I would expand this out to make it clear.

herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_worker_.m
```diff
@@ -9,6 +9,14 @@
 end
 scr_path = which(new_wrkr);
 if isempty(scr_path)
+    % Check if it is a compiled worker
+    new_wrkr = check_compiled_(new_wrkr);
+    if ~isempty(new_wrkr)
+        config_store.instance().store_config(obj, 'worker', new_wrkr);
```
oerc0122, line 15: `config_store.instance()` is used a lot here, does this want to be instanced and assigned to a variable to avoid repetition?

herbert_core/classes/MPIFramework/@parallel_config/parallel_config.m
```diff
@@ -215,22 +215,19 @@
         %-----------------------------------------------------------------
         % overloaded getters
         function wrkr = get.worker(obj)
-            wrkr= get_or_restore_field(obj,'worker');
+            wrkr = get_or_restore_field(obj, 'worker');
```
oerc0122, line 218: Is this a method and thus does this want to be called like a method?
`obj.get_or_restore_field(...)`

herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
```diff
@@ -172,9 +172,20 @@ function check_availability(obj)
...
-                % future.
-                mpi_exec = fullfile(external_dll_dir, 'mpiexec.exe');
+                [rs, rv] = system('where mpiexec');
+                if rs == 0
+                    mpi_exec = strip(rv);
+                    mpis = splitlines(mpi_exec);
+                    if numel(mpis) > 1
+                        % If multiple mpiexec on path, prefer user installed MS MPI
+                        mpi_id = find(cellfun(@(x) ~isempty(strfind(x,'Microsoft')), mpis), 1);
+                        if isempty(mpi_id), mpi_id = 1; end
```
oerc0122, line 182: I personally dislike inline-`if`s, expand out?

herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
```diff
@@ -172,9 +172,20 @@ function check_availability(obj)
             rootpath = fileparts(which('herbert_init'));
             external_dll_dir = fullfile(rootpath, 'DLL','external');
             if ispc()
-                % only one version of mpiexec is used now. May change in the
-                % future.
-                mpi_exec = fullfile(external_dll_dir, 'mpiexec.exe');
+                [rs, rv] = system('where mpiexec');
```
oerc0122, line 175: Do we want the equivalent for where `ispc` is false?

oerc0122 (approved):
If this works that's good, otherwise, just some minor formatting complaints.

herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_cluster_.m
mducle: yeah... maybe... it's just it's in three different classes... :/

herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_worker_.m
mducle: What we really need is a ternary operator...

abuts:
@mducle Do you really want it to rel_3_6? Ok, tests for master will probably fail then.

mducle:
> @mducle Do you really want it to rel_3_6? Ok, tests for master will probably fail then.

Well... the changes are needed for `pace-python` to work. At the moment I'm [patching](https://github.com/pace-neutrons/pace-python/blob/main/src/CMakeLists.txt#L46) the 3.6.2 sources but I would like to be able to use it in directly without patching in e.g. 3.6.3.

I guess I could apply this PR to `master` and then back-port it to `rel_3_6` but I would like it to be in the release...

abuts (approved):
Given sticky points, I have mentioned, work, looks ok.  Have not noticed changes to `ClusterSlurm` 
an `ClusterHerbert` are they all use the same launcher as `ClusterMpi`?  I though each one has its own, but may be I am wrong and the launcher is placed on the parent. 

Then its fine.

abuts:
Sorry, I was writing my comments alongside with Jacob's but the editor separated them. They may look a bit unclear separately.

abuts:
> I guess I could apply this PR to `master` and then back-port it to `rel_3_6` but I would like it to be in the release...

You started it from rel_3_6, so you merging it to 3.6 is ok. Then branch should be relatively easy merged to master. This actually makes me wander, why the tests are failing. ...

herbert_core/classes/MPIFramework/@ClusterParpoolWrapper/ClusterParpoolWrapper.m
```diff
@@ -137,7 +137,11 @@
             % variables, but if the cluster is remote, the envriomental
             % variables transfer should be investigated
             obj.set_env();
-            h_worker = str2func(obj.worker_name_);
+            if obj.is_compiled_script_
+                h_worker = @worker_v2;
+            else
```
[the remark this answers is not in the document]
mducle: Right... yes, I had misunderstood how it worked... so if you compile Matlab with the parallel toolbox it will spawn additional processes for `parpool` without needing a compiled worker... which is why it need the reference to the worker function. I'll remove this and make it an error (via an `assert` to initialise a `@ClusterParpoolWrapper` with a compiled worker).

herbert_core/classes/MPIFramework/@parallel_config/parallel_config.m
```diff
@@ -215,22 +215,19 @@
...
         end
         function wrkr = get.is_compiled(obj)
-            % incomplete! Should be derived from worker
-            wrkr= obj.is_compiled_;
+            wrkr = get_or_restore_field(obj, 'is_compiled');
```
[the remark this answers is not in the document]
mducle: Yeah, this is just the `get` function - there is no `set` for `is_compiled` it can only be set by setting `worker`. 

If you set `worker` to something which is found by _Matlab_'s `which` (implying it is a Matlab function) then `is_compiled` is `false`. If you set `worker` to something which is not found by _Matlab_'s `which` but is a fully qualified path to an existing file, _or_ is something found by the _system_ `which` (`where` on Windows) then `is_compiled` is `true`.

herbert_core/classes/MPIFramework/@ClusterMPI/ClusterMPI.m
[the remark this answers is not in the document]
mducle: I tested it with [MSMPI v8](https://www.microsoft.com/en-us/download/confirmation.aspx?id=55494), [MSMPI v9](https://www.microsoft.com/en-us/download/details.aspx?id=56727) and [MSMPI v10](https://www.microsoft.com/en-us/download/details.aspx?id=100593) and it works on all 3 with the Jenkins compiled `mex` files - I just download the release bundle and use it as is; I don't recompile the `mex` files.

I think that the `cpp_communicator` only uses MPI2.2 (or earlier) features and that standard came out in 2009 and is supported by all three versions of MS MPI we're talking about here. It seems there are no ABI changes between the versions either so they all seem to work ok.

The issue with the Horace-bundled `mpiexec.exe` is due to the firewall (at least for me; I think). Even though I set it to be allowed by the firewall it keeps failing with a `Failed to post close command error 1726`... it might be due to some signing issues but I couldn't figure out a way around it except by installing the official package...

herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_worker_.m
```diff
@@ -32,6 +40,23 @@
...
+    config_store.instance().store_config(obj, 'is_compiled', false);
 end
+end % function
 
+function out = check_compiled_(worker)
+    out = '';
+    if exist(worker, 'file')
+        % Assume if input is full path to file, then it is a compiled worker
+        out = worker;
+    else
```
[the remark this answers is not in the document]
mducle: It should be fine because this check only occurs if the `which(worker)` check yields empty. But I'll add a `&& ~endsWith(worker, '.m')` to the `if`.

herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_worker_.m
```diff
@@ -32,6 +40,23 @@
...
+        if ispc(), cmd = 'where'; else, cmd = 'which'; end
+        [rs, rv] = system([cmd ' ' worker]);
+        if rs == 0
+            % Assume if it is on the system path, then it is a compiled worker
+            out = splitlines(strip(rv));
+            out = out{1}; % Only take first path if there is more than one
+        end
```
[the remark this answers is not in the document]
mducle: So this bit is to allow a user shortcut because what we want is the fully qualified path to pass to the Java `ProcessBuilder`, so if the user just sets `pc.worker = 'horace_worker.exe'` without the full path it would not work; so the `system('[which|where] <worker>')` here is to get the full path. `.m` files will not be returned by this (unless they have erroneously been marked as executable by the filesystem on Linux).

herbert_core/classes/MPIFramework/@parallel_config/private/check_and_set_worker_.m
mducle: it still works...

PR #443, opened 2022-05-06
Herbert442 unique objects container

cmarooney-stfc:
Implements a container `unique_objects_container` which appears to store N objects {ai==a1,a2....aN} and can retrieve them using index i, but actually stores only the non-duplicate objects. Index is internally mapped to the indices of the cell array internally storing the objects.

Based on the design by Duc in Horace PR 790.

The container compares objects by creating an MD5 hash for each of them. Default is to use the undocumented stream function seen in the code. Alternative streams such as hlp_serialise can be specified in the constructor.

The stored objects can be limited to sub-classes of a class specified in the constructor.

A test suite is also supplied.


abuts (approved):
looks ok if works and it looks like it works. 

Substantial issue:
Why you want to have container as a handle? You would have at least 3 different containers. May be 4 in experiment and I do not see any reason of sharing them. Two independent sqw objects would try to share their containers and you would need special measures to avoid this. 

I am getting ready to kill handle from pixels (but there may be justification for it there, as a file accessor is a handle) but why here? why can not two independent containers exist?

No reason to have it as a handle. 

Minor remarks:

2) I would use `ismember(hash, existing_hashes)` function to check belonging to the array of hashes. It is much simple and `pos = ismember(existing_hashes,hash)` would return the position of the object in hash, where operation obj = unique_obj_storage{pos} would be effective way of retrieving the requested object.

3) should java `Engine` be a persistent property ,initialized on the first access?  Looks like building it each time is expensive. 

4) would it be nice to overload {} or () to extract object from container by indexes instead of using `get`?

Otherwise looks acceptable. At least all further development can be easy build on it.

mducle (commented):
Looks good.

As Alex says, considering that `sqw` are value classes then assigning `w2 = w1` where `w1` is an `sqw` (or `dnd`) would not copy the unique instrument containers... so this would then be shared between the two (in principle separate copies) `w1` and `w2`. We can make the unique container a value class instead but this means using `uoc = uoc.add(li)` instead of just `uoc.add(li)` (or `[uoc, nuix] = uoc.add(li)` instead of `nuix = uoc.add(li)`) and changing the `add` method to return `self`.

From an aesthetic point of view I don't like this (just like I don't like the `multifit`/`tobyfit` syntax), and if we are moving towards Python we should really be translating everything to `handle` instead...

The alternative is to make some sort of hook so that when `w2 = w1` is invoked a copy of the container is created... but this could be quite messy and bug-prone. So, in terms of making everything work together in `sqw` / `dnd` this unique container should probably be a value classs instead of handle... (even if it is ugly...)

PS. I thought that we had a discussion a couple of years ago and agreed to move everything to `handle`.... I admit that I've not been paying much attention though...

PR #444, opened 2022-05-13
extend unique_objects_container design

cmarooney-stfc:
unique_objects_container has been refactored and extended in response to comments and
design needs revealed by testing. Changes are:

1. Container is now not a handle class (it was not intended to be, may have got in from Duc's
original design.) This also means that methods changing the object must return self.
2. Container is now serializable - needed to serialise Experiment class. 
3. Container can now either emulate an array or a cell array - specified by constructor parameter.
3. Subscripted container can return an element value (subsref) and can set an element value.
For setting, the next element above the container size can be set, and an element already in the
container can be replaced. In the latter case, the case in which a unique object no longer has any
non-unique object indexes is tracked and reported, but at present no action is taken to error or remedy
the situation.

See Herbert PR 442 for initial design


abuts (changes requested):
looks fine on first glance (will make a second one) but the tests are not included neither in Jenkins (cmake lists) not into manual testing list (validate_herbert)

abuts (commented):
I am surprised that internal properties, assuring container integrity, like duplicated indexes and duplicated objects (and all others) have external setters. 
Why? if you set it up externally, you will violate the container integrity. Should not be there.

PR #452, opened 2024-02-13
possible fixture for issues with inalid configurations when Horace3 switched to Horace4

abuts:


abuts:
this is invalid PR erroneusly made agains master rather then against rel_3_6_3

PR #453, opened 2024-02-13
Possible fixture for problem when people switching from Horace3 to Horace4

abuts:
Some people have old startup file which contans direct call to herbert_init, Horace_init. 

When people with such config switch configurations, they may have problem as herbert3 remains on the path. 
To fix such issues, suggested modifications warn them that they have such startup in their path. 
old `herbert_init` moved to `herbert3_init` and `herbert_init` calls `herbert3_init` and issues warning


herbert_core/admin/herbert_version.m
```diff
@@ -1,4 +1,4 @@
-function varargout = herbert_version()
+function varargout = herbert_version(varargin)
```
cmarooney-stfc, line 1: may be worth adding non-empty varargin examples to the usage section. But I don't see any usage of varargin - why are you providing it?

herbert_core/herbert3_init.m
```diff
@@ -0,0 +1,144 @@
+function herbert3_init
+% Adds the paths needed by Herbert.
+%
+% Do not call this script directly. Use horace_on or herbert_on 
+% command which enables this script
+% 
+%
+% Is PC and Unix compatible.
+
+% T.G.Perring
```
cmarooney-stfc, line 10: I think this is your authorship, not Toby's, even if you have cut/pasted it from his code

cmarooney-stfc (changes requested):

herbert_core/admin/herbert_version.m
abuts: new version of `herbert_version` accepts arguments. As it may be unclear which version is used during initialization or autocheck for version may be performed I have added the argument to old version to maintan the same interface.
Do not see need for providing usage for the interface yet but may be necessary in a future

herbert_core/herbert3_init.m
abuts: woud disagree -- I've fully copied Toby's code into function with new name. If he wants to claim authority -- it is certainly his code.

